PR #1, opened 2016-06-27
Refactor the ImageTransform class

rowanmanning:
Refactoring the `ImageTransform` class to stop using getters/setters. This should be a bit clearer/easier to understand now.


AlbertoElias:
looks good!

PR #8, opened 2016-06-28
Flesh out the service endpoint documentation

### LINKED ISSUE #2: Document how to use the image service
This can be mostly copied from the [existing image service](http://image.webservices.ft.com/v1/), and added to [`docs/index.html`](https://github.com/Financial-Times/origami-image-service/blob/master/docs/index.html).

rowanmanning:
This resolves #2. A lot of this is copied from/based on the existing Image Service documentation.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -43,6 +43,15 @@
 		a {
 			color: #2e6e9e;
 		}
+		td {
+			vertical-align: top;
+		}
+		td p:first-child {
+			margin-top: 0;
+		}
+		dt {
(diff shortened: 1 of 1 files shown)


AlbertoElias:
Only thing is that it mentions it, but doesn´t have any examples of using it with our custom image sets

rowanmanning:
Oh it shouldn't mention it, I thought I removed all references. That's not currently built

AlbertoElias:
oh, ok! good to go then

PR #9, opened 2016-06-29
Add a proxy response handler to tidy up headers

rowanmanning:
The headers that come back from Cloudinary and Imgix are a bit
inconsitent, and they include a lot of stuff we don't really want to
send as part of the response.

This removes them and sets consistent headers no matter where the image
comes from.


AlbertoElias:
lgtm!

PR #10, opened 2016-06-30
N express

tavvy:
requires a change to n-express:
 https://github.com/Financial-Times/n-express/blob/master/main.js#L183 needs to `return`


rowanmanning:
@tavvy this looks great 👍 is it ready to merge?

tavvy:
👍

Issue #11, opened 2016-07-11
Add in support for custom schemes


rowanmanning:
We need to support the custom schemes that version 1 supports: `ftcms`, `fticon`, `fthead`, `ftsocial`, `ftpodcast`, `ftlogo`.

These map to various Git repositories, and there are a couple of options for doing this:

1. On deploy, all of the required repositories are installed on the server and custom schemes map to the created folder of images. This is similar to way v1 of the Image Service works. This method is complicated by the fact that we need to apply transforms through proxying to a third-party – we'd need to publicly expose the raw images and then hit the image service twice to load them:

  ```
  > https://image-service/v2/images/raw/fticon:cross?format=png
  [ expands to ]
  > https://image-service/v2/images/raw/https%3A%2F%2Fimage-service%2Fv2%2Fimages%2Fsource%2Ffticon%2Fcross.svg?format=png
  [ proxies to ]
  > https://third-party/image/https%3A%2F%2Fimage-service%2Fv2%2Fimages%2Fsource%2Ffticon%2Fcross.svg?format=png
  [ third-party hits the image service again to request the icon ]
  ```

2. On request, the URL is mapped to a _different_ publicly hosted version of the images. This is very similar to option 1 except that the image service itself wouldn't be responsible for serving the images as well as proxying them:

  ```
  > https://image-service/v2/images/raw/fticon:cross?format=png
  [ expands to ]
  > https://image-service/v2/images/raw/https%3A%2F%2Fother-service%2Ffticon%2Fcross.svg?format=png
  [ proxies to ]
  > https://third-party/image/https%3A%2F%2Fother-service%2Ffticon%2Fcross.svg?format=png
  [ third-party hits the other service to request the icon ]
  ```

3. On request, the URL is mapped to an image that has been uploaded to the third party already. This would require us to build an image uploader as [outlined in the original proposal](https://docs.google.com/drawings/d/1fC07fJ_2Flwd5Qugyltb_5mU2ALqtssaPqMdy_5-xN4). This ties us further into a specific third party and could be complicated to build, but it offers the simplest/most efficient request process:

  ```
  > https://image-service/v2/images/raw/fticon:cross?format=png
  [ proxies to ]
  > https://third-party/uploaded/fticon/cross?format=png
  ```

As well as reimplementing what we already have in Image Service v1, we also need to allow versioning of image sets. Currently we have no ability to remove images as it would result in dependent sites breaking. We'd like to support a version in the custom scheme like this:

```
[...]

PR #16, opened 2016-07-13
Add support for custom schemes

rowanmanning:
See #11 for more information.


lib/image-transform.js
```diff
@@ -156,6 +158,54 @@ module.exports = class ImageTransform {
...
+		baseUrl = baseUrl.replace(/\/+$/, '');
+		pathname = pathname.replace(/\/+$/, '');
+
+		let version;
+		if (/\-v\d+$/.test(scheme)) {
+			scheme = scheme.split('-');
+			version = scheme.pop();
+			scheme = scheme.join('-');
+		}
+		version = version || 'unversioned';
```
AlbertoElias: I find hard to follow all the transformations and info extraction that's going on here, could it possible have some more comments?

lib/routes/v2/images-debug.js
```diff
@@ -1,15 +1,34 @@
...
+				transform: request.transform,
+				appliedTransform: request.appliedTransform
+			});
+		}
+	);
+
+	// Image with a custom scheme
+	// TODO make sure these default to the original exension! E.g. SVG for fticon
+	app.get(
+		/^\/v2\/images\/debug\/((ftcms|fticon|fthead|ftsocial|ftpodcast|ftlogo)(\-v\d+)?(:|%3A).*)$/,
```
AlbertoElias: same for regexps, there are quite a few

lib/image-transform.js
rowanmanning: Yeah absolutely. Sorry, I've got out of the habit which is bad!

test/unit/lib/image-transform.js
```diff
@@ -497,6 +505,213 @@ describe('lib/image-transform', () => {
...
+
+			it('throws an error', () => {
+				assert.throws(() => {
+					ImageTransform.resolveCustomSchemeUri('foo:example', 'http://base/images');
+				}, 'Image URI must be a string with a valid scheme');
+			});
+
+		});
+
+		describe('when `uri` doesn\'t have a hostname', () => {
```
AlbertoElias: The error is the same as above, and why is icon name called hostname?

rowanmanning: Yeah I'm making sure that the error gets triggered properly on different cases. It's called a hostname in the test because when you parse it as a URL the icon name comes through as `parsedUrl.hostname`. I can change if it's confusing

lib/image-transform.js
```diff
@@ -158,45 +266,71 @@ module.exports = class ImageTransform {
...
 			case 'fticon':
 			case 'ftlogo':
 			case 'ftsocial':
 				if(!path.extname(pathname)) {
 					pathname = `${pathname}.svg`;
 				}
 				return `${baseUrl}/${scheme}/${version}/${pathname}`;
+
+			// PNG-based schemes. Here we're mostly adding
+			// the .svg extension if there isn't an extension
```
AlbertoElias: .png*

AlbertoElias:
left one comment, rest looks good!

PR #17, opened 2016-07-21
Handle the ftcms custom scheme

rowanmanning:
This mirrors the behaviour of the original image service, first
attempting to load an image from the V1 API and then trying V2.

In order to utilise the proxying behaviour of this service we
don't load the image, we make a HEAD request to verify that an
image is available.

If neither can be found then we error.

### DIFF
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -29,7 +29,7 @@ function getCmsUrl() {
 				return fetch(v2Uri, fetchOptions);
 			})
 			.then(secondResponse => {
-				// Cool, we've got an image from v2
+				// Cool, we've got an image from v2 (or v1 if that is OK)
 				if (secondResponse.ok) {
 					return secondResponse;
 				}
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -1,0 +1,45 @@
+'use strict';
+
+const httpError = require('http-errors');
+
+module.exports = getCmsUrl;
+
+function getCmsUrl() {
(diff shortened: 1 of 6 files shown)


test/unit/lib/middleware/get-cms-url.js
```diff
@@ -0,0 +1,117 @@
...
+
+		it('returns a middleware function', () => {
+			assert.isFunction(middleware);
+		});
+
+		describe('middleware(request, response, next)', () => {
+			let mockFetchResponseV1;
+			let mockFetchResponseV2;
+
+			beforeEach(done => {
```
AlbertoElias: should this be a before or a beforeEach?

lib/middleware/get-cms-url.js
```diff
@@ -0,0 +1,45 @@
...
+				// Cool, we've got an image from v1
+				if (firstResponse.ok) {
+					return firstResponse;
+				}
+				// If the v1 image can't be found, try v2
+				return fetch(v2Uri, fetchOptions);
+			})
+			.then(secondResponse => {
+				// Cool, we've got an image from v2
+				if (secondResponse.ok) {
```
AlbertoElias: this would be equal to firstResponse if that one was ok, right?

rowanmanning: Yep. I thought it'd be nicer to do it like this than try to nest promises.

rowanmanning: I could just name them all fetchResponse? Would that be clearer?

AlbertoElias:
Good to go!

Issue #12, opened 2016-07-11
Add in support for SVG tinting


rowanmanning:
We have a few options for this:

1. _Don't_ add support for SVG tinting. Instead we would add build scripts to each of the repositories that have SVGs in them, creating SVGs in all of the allowed colours. They would then be references using something like `fticon:cross/pink`.

2. Try to use a third party. This won't be possible with Imgix, but Cloudinary is adding a way to tint images in the same way as our v1 Image Service does. If this supports SVGs in the same way then we could rely on that.

3. Post-process SVGs in the new image service, adding in the style block to tint them. This adds a layer of complexity to the service which we might not want, but it's likely the closest we can get to the existing behaviour.

PR #20, opened 2016-08-08
Implement an SVG tinting endpoint

rowanmanning:
This adds partial support for #12. The next part of this work is routing any SVG image requests through this additional layer if they have a tint parameter.


AlbertoElias:
Only issue is current image service accepts 2 colours

rowanmanning:
Yep, I should add a comment explaining how we deal with it.

AlbertoElias:
lgtm

PR #21, opened 2016-08-09
Route tinted SVGs through the svgtint endpoint

rowanmanning:
This doesn't have tests yet because I want to see how it works on qa – third party services don't have access to my local server and so the routing doesn't work properly. I'll be opening a second PR with automated tests once I know this works.


AlbertoElias:
:+1:

PR #22, opened 2016-08-10
Fix SVG header sending and write tests

rowanmanning:
SVG headers weren't being set properly, and a format of "auto" would never result in SVGs being sent. This fixes it and adds some tests to prevent it from happening again.


AlbertoElias:
lgtm

PR #23, opened 2016-08-10
Move SVG tinting into middleware and write tests

rowanmanning:
This introduces unit tests for the SVG tinting that was added in #20. It also moves all of that code into reusable (and more testable) middleware.


AlbertoElias:
👍

PR #24, opened 2016-08-11
Move tinting validation logic into ImageTransform

rowanmanning:
It's more testable here and allows us to use `ImageTransform`'s colour name conversion stuff. E.g. `red` to `#ff0000`.


AlbertoElias:
👍

PR #25, opened 2016-08-15
Remove Docker support

rowanmanning:
This is the same work as was done on the Build Service,
Heroku's Docker support isn't _quite_ ready for us to use
in production so we're reverting to slug-based builds for
the moment.


AlbertoElias:
looks good

PR #26, opened 2016-08-16
Default the tint parameter and document it

rowanmanning:
This brings the tint parameter more in line with v1 of the service and documents how it behaves.

### DIFF
--- a/docs/index.html
+++ b/docs/index.html
@@ -217,7 +217,7 @@
 							<td><code>tint</code></td>
 							<td>Querystring</td>
 							<td>
-								<p>Replaces colours in the image. The exact behavior depends on the number of comma-separated colours specified, and colours can be either hex codes or named colours. Tinting only works when the source image is an SVG.</p>
+								<p>Replaces colours in the image. Tinting only works when the source image is an SVG, and colours can be either hex codes or named colours. Different numbers of colours (separated by commas) behave differently:</p>
 								<ol start="0">
 									<li>The <code>tint</code> parameter used with no value (no colours given) tints the image with FT's "pink" colour.</li>
 									<li>When given one colour, e.g. <code>fff1e0</code>, it tints the image with the given colour.</li>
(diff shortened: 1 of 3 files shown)


AlbertoElias:
PR looks good. Are we limiting the colours used when tinting to colours in o-colors?

rowanmanning:
We're not at the moment, as I wanted to maintain backwards compatibility as much as possible with existing image URLs.

docs/index.html
```diff
@@ -213,6 +213,18 @@ <h2 id="reference">API reference</h2>
 								<p>Fallback background colour to apply when the output image format does not support transparency (i.e., JPEG). Ignored if the image supports transparency. Specified as three- or six-character RGB hex code, e.g. <code>00ff00</code></p>
 							</td>
 						</tr>
+						<tr id="tint">
+							<td><code>tint</code></td>
+							<td>Querystring</td>
+							<td>
+								<p>Replaces colours in the image. The exact behavior depends on the number of comma-separated colours specified, and colours can be either hex codes or named colours. Tinting only works when the source image is an SVG.</p>
```
onishiweb: I think this reads a bit odd. It's confusing that you mention the behaviour depends on the colours specified, then mention that when two or more colours are specified it will default to the first and then says that it's deprecated.

rowanmanning: Yeah I see what you mean. Do you have any suggestions on how to improve? It was largely copied over from the existing documentation

onishiweb: I'd maybe remove the `The exact behavior depends on the number of comma-separated colours specified` and replace it switch around the sentences to put the important information first perhaps, so it would read:

`Tinting only works when the source image is an SVG, and colours can be either hex codes or named colours`.

docs/index.html
```diff
@@ -213,6 +213,18 @@ <h2 id="reference">API reference</h2>
...
+								<ol start="0">
+									<li>The <code>tint</code> parameter used with no value (no colours given) tints the image with FT's "pink" colour.</li>
+									<li>When given one colour, e.g. <code>fff1e0</code>, it tints the image with the given colour.</li>
+									<li>When given two or more colours, e.g. <code>fff1e0,ff0000</code>, it tints the image with the first colour. This behaviour is in place to reduce work when migrating between versions 1 and 2 of the image service, and should be considered deprecated.</li>
```
onishiweb: If you're making the above change, I'd perhaps update the last bullet point to put a bit more emphasis on the deprecation, maybe just add `(deprecated)` to the start to make it clearer?

PR #30, opened 2016-08-19
Map unversioned schemes to specific image set versions

### LINKED ISSUE #29: Explicitly map unversioned schemes
Currently unversioned schemes point to an `unversioned` folder, this needs to change to have an explicit version set. This will probably be best as a map of scheme to version. We also need this to be able to rewrite the scheme if necessary to support the old icon set. E.g.

```js
const unversionedSchemeMap = {
    fticon: 'fticon-old/v4', // this has a scheme and version so will rewrite the scheme
    fthead: 'v4'             // this only has a version so the fthead scheme will be used
};
```

rowanmanning:
Resolves #29.

### DIFF
--- a/lib/image-transform.js
+++ b/lib/image-transform.js
@@ -346,8 +346,20 @@ module.exports = class ImageTransform {
 			version = scheme.pop();
 			scheme = scheme.join('-');
 		}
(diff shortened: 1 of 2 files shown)


AlbertoElias:
👍

Issue #27, opened 2016-08-17
Anonymise the request before proxying


rowanmanning:

JakeChampion:
We should find out what Headers our image-optimising provider requires to function correctly and only send those. This can be done as a whitelist, filtering out any other headers.

PR #31, opened 2016-08-30
Anonymise user requests before proxying to a third party

rowanmanning:
This attempts to address #27 by removing potential personal data from the request before proxying it to a third party. Currently we're setting the `User-Agent` header to the image service, and then removing cookies.

Are there any other common headers which might contain information that could identify the end user? I feel like I've probably missed a lot.

### DIFF
--- a/lib/image-service.js
+++ b/lib/image-service.js
@@ -6,6 +6,7 @@ const httpProxy = require('http-proxy');
 const morgan = require('morgan');
 const notFound = require('./middleware/not-found');
 const oneWeek = 60 * 60 * 24 * 7;
+const pkg = require('../package.json');
 const requireAll = require('require-all');
 const url = require('url');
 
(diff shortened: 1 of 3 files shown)


lib/image-service.js
```diff
@@ -46,6 +47,9 @@ function createProxy(errorHandler) {
 
 	function proxyRequestHandler(proxyRequest, request, response, proxyOptions) {
 		proxyRequest.setHeader('Host', url.parse(proxyOptions.target).host);
+		proxyRequest.setHeader('User-Agent', `${pkg.name}/${pkg.version}`);
```
JakeChampion: We should find out what Headers our image-optimising provider requires to function correctly and only send those. This can be done as a whitelist, filtering out any other headers.

rowanmanning: Good idea, I've pushed some changes and this seems to work. What do you think?

JakeChampion:
Great work here @rowanmanning - merging this in :-)

Issue #33, opened 2016-08-31
Add healthchecks


rowanmanning:
The `__health` endpoint currently has no healthchecks. These need to be decided on and added.

alicebartlett:
- availability of cloudinary
- availability of s3 bucket where image sets are stored
- memory

rowanmanning:
cloudinary/s3 checks are in place, memory is being recorded in StatsD at the moment.

PR #38, opened 2016-09-05
Add health-checks for Cloudinary and S3

rowanmanning:
This gets us some of the way towards completing #33, adding in health-checks for Cloudinary and the S3 bucket we use to store image sets.

### DIFF
--- a/lib/health-checks.js
+++ b/lib/health-checks.js
@@ -1,0 +1,71 @@
+'use strict';
+
+const cloudinary = require('cloudinary');
+const pingInterval = 60 * 1000;
+
+let customSchemeCheckUrl;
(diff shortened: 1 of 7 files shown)


lib/health-checks.js
```diff
@@ -0,0 +1,66 @@
...
+const pingInterval = 60 * 1000;
+const serviceStatuses = {};
+
+let customSchemeCheckUrl;
+let cloudinaryCheckUrl;
+
+// Health checks
+module.exports = {
+
+	init(config) {
```
JakeChampion: We should add some tests for the healthchecks.

JakeChampion:
LGTM 👍

PR #42, opened 2016-09-07
"__compare" endpoint to compare between versions

mariabg:


data/comparison-images.json
```diff
@@ -0,0 +1,37 @@
+[ "/images/raw/http%3A%2F%2Fim.ft-static.com%2Fcontent%2Fimages%2Fa60ae24b-b87f-439c-bf1b-6e54946b4cf2.img?source=test",
```
rowanmanning: Could we tab-indent this file?

lib/routes/__compare.js
```diff
@@ -0,0 +1,20 @@
+'use strict';
+
+// Generate image variants
+const imagesSrc = require('../../data/comparison-images.json');
+
+// imagesSrc.forEach((img) => {
+// 	return img;
+// });
```
rowanmanning: Commented code should be removed :)

lib/routes/__compare.js
```diff
@@ -0,0 +1,20 @@
...
+
+// original stuff
+module.exports = app => {
+	app.get('/__compare', (request, response) => {
```
rowanmanning: I don't think we need the double underscores here. I'd be happy if it was just `/compare`, as it'd be considered a public URL and we'll probably link to it

rowanmanning:
I made a couple of comments but this looks great. Really useful to have 👍

rowanmanning:
Actually running locally it's odd, the table cells aren't the same widths and it means that the images are scaling differently. It might be good to explicitly set the table cells to be 50% wide?

mariabg:
I made some changes, please checkout if they are what you had in mind :)

rowanmanning:
Looks great 👍 thanks!

mariabg:
Merge?

PR #43, opened 2016-09-08
Disable tinting of SVG strokes for compatibility

rowanmanning:
This is for backwards compatibility with v1 of the image service. Some SVGs break slightly if we tint the stroke colour too.


JakeChampion:
LGTM 👍

Issue #13, opened 2016-07-11
Ensure that the DPR transform uses client hints


rowanmanning:
The current Image Service sends a `content-dpr` header when the DPR is set. Investigate why it does this, and ideally make sure that v2 of the Image Service works in the same way.

mariabg:
#51

PR #44, opened 2016-09-08
DPR transformation working

mariabg:
#13

### DIFF
--- a/lib/image-transform.js
+++ b/lib/image-transform.js
@@ -23,9 +23,9 @@ module.exports = class ImageTransform {
 	 */
 	constructor(properties) {
 		this.setUri(properties.uri);
-		this.setWidth(properties.width, properties.dpr);
-		this.setHeight(properties.height, properties.dpr);
-		this.setDpr(properties.dpr);
+		this.setWidth(properties.width);
+		this.setHeight(properties.height);
+		this.setDpr(properties.dpr, properties.width, properties.height);
 		this.setFit(properties.fit);
 		this.setQuality(properties.quality);
 		this.setFormat(properties.format);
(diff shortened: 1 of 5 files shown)


lib/image-transform.js
```diff
@@ -96,9 +96,20 @@ module.exports = class ImageTransform {
...
 		this.dpr = ImageTransform.sanitizeNumericValue(value, errorMessage);
+		if(this.dpr > 1) {
+			if (this.height) {
+				this.setHeight(Math.floor(height / this.dpr));
+			}
+			if (this.width) {
+				this.setWidth(Math.floor(width / this.dpr));
+			}
+			if(!this.width && !this.height) {
+				throw new Error (errorMessage);
```
rowanmanning: I think should throw a specific error. Something like:
```
Image DPR can only be set when the transform has an explicit width or height
```

views/index.html
```diff
@@ -97,6 +97,7 @@ <h2 id="reference">API reference</h2>
 						<td>
 							<p>Device Pixel Ratio, i.e. number of image pixels per CSS pixel. The default is <code>1</code>, which gives the standard 1:1 ratio of image pixels to CSS pixels. Set to <code>2</code> for high-DPI ("Retina") displays. Request to the image service with <code>width=100<wbr>&amp;dpr=2</code> will output image 200 pixels wide, but optimized to be displayed with CSS <code>width:100px</code></p>
 							<p>Note that browsers don't automatically adjust intrinsic size of images with higher DPI, so you need'll need to use some "responsive image" solution, e.g. set explicit width or height, use <code>image-set()</code> or <code>srcset</code> attribute.</p>
+							<p>Take also into account that if this parameter is set to a value <code>>1</code> and the <code>height</code> or <code>width</code> of the image is not specify, it will error</p>
```
rowanmanning: > ...the image is not **specify**, it will error

should be

> ...the image is not **specified**, it will error

rowanmanning:
Couple of comments, otherwise looking great. Thanks 👍

lib/image-transform.js
```diff
@@ -96,9 +96,20 @@ module.exports = class ImageTransform {
...
 	 */
-	setDpr(value) {
+	setDpr(value, width, height) {
 		const errorMessage = 'Image DPR must be a positive whole number';
 		this.dpr = ImageTransform.sanitizeNumericValue(value, errorMessage);
+		if(this.dpr > 1) {
+			if (this.height) {
+				this.setHeight(Math.floor(height / this.dpr));
+			}
+			if (this.width) {
```
rowanmanning: Oh actually these conditionals should be checking for the `width` and `height` parameters, not properties on the instance.

Issue #47, opened 2016-09-13
Make the error pages more useful


rowanmanning:
The error pages in v2 aren't as useful or well styled as v1. We should be able to use the existing Handlebars templates to spruce these up and also make sure all the errors we send are clear.

rowanmanning:
We haven't ever prioritised this, and we don't get any complaints. Mostly error pages aren't seen by users of the service because browsers will display as a broken image

PR #48, opened 2016-09-13
Render the error pages with Handlebars

rowanmanning:
Making a start on #47.


JakeChampion:
Looks like a good start Rowan 👍

PR #49, opened 2016-09-13
Redirect v2

rowanmanning:
- **Switch to `501 Not Implemented` for `/v1` URLs:**  
   I think this makes more sense than `410 Gone`, as the resource will never have existed
- **Redirect the application index page to /v2:**  
   This makes more sense than redirecting to a `501` :)

### DIFF
--- a/test/integration/home.js
+++ b/test/integration/home.js
@@ -8,8 +8,8 @@ describe('GET /', function() {
 	setupRequest('GET', '/');
 	itRespondsWithStatus(301);
 
-	it('redirects the request to the v1 documentation page', function(done) {
-		this.request.expect('Location', '/v1/').end(done);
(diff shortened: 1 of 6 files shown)


test/integration/home.js
```diff
@@ -8,8 +8,8 @@ describe('GET /', function() {
 	setupRequest('GET', '/');
 	itRespondsWithStatus(301);
 
-	it('redirects the request to the v1 documentation page', function(done) {
-		this.request.expect('Location', '/v1/').end(done);
+	it('redirects the request to the v2 documentation page', function(done) {
+		this.request.expect('Location', '/v2/').end(done);
```
JakeChampion: :+1

JakeChampion:
LGTM 👍

PR #52, opened 2016-09-19
Finish the documentation

rowanmanning:
This gets us to the point where the service is documented enough for a public beta.


lib/routes/v2/docs/api.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+const oneWeek = 60 * 60 * 24 * 7;
+
+module.exports = app => {
+
+	// v2 api documentation page
+	app.get('/v2/docs/api', (request, response) => {
```
JakeChampion: We could use composed express routers to avoid having to write `/v2/doc` in these route handlers.

rowanmanning: I agree. I kind of want to rethink routing as a whole though, I think we could reduce some of the boilerplate :)

PR #53, opened 2016-09-19
Allow setting a base URL

rowanmanning:
This allows us to run the Image Service on a sub-path rather than
immediately after the hostname.

### DIFF
--- a/lib/image-service.js
+++ b/lib/image-service.js
@@ -19,6 +19,8 @@ function imageService(config) {
 	const app = createExpressApp(config);
 	const errorHandler = handleErrors(config);
 	app.proxy = createProxy(errorHandler);
+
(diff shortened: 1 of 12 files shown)


lib/image-service.js
```diff
@@ -110,3 +112,13 @@ function mountRoutes(app) {
 		resolve: initRoute => initRoute(app)
 	});
 }
+
+function sanitizeBaseUrl(url) {
```
JakeChampion: Hmm, didn't think of that.

rowanmanning: Alright to merge in that case?

rowanmanning:
@JakeChampion actually after thinking about it on the way into work I think this solution is the best one to go with for now. I don't think it's a _huge_ amount of work to overhaul it, but this is fine for MVP. I've opened an issue to revisit this later (#54). Am I OK to merge?

JakeChampion:
Yep :)

PR #56, opened 2016-09-20
Remove the baseUrl configuration

rowanmanning:


lib/image-service.js
```diff
@@ -20,7 +20,8 @@ function imageService(config) {
 	const errorHandler = handleErrors(config);
 	app.proxy = createProxy(errorHandler);
 
-	config.baseUrl = app.locals.baseUrl = sanitizeBaseUrl(config.baseUrl);
+	config.basePath = config.basePath || '';
```
JakeChampion: Isn't this always set? When would we hit the `''` case?

rowanmanning: It hits this case in the tests, where we don't use `config.js`

PR #84, opened 2016-10-17
Comparison sizes

rowanmanning:
Add image formats and sizes to the comparison page. The v1 URLs are static and set in the comparison JSON; the v2 URLs are fetched on page load so that we can inspect the response information.

This allows us to manually (for now) ensure that changes we make do not result in a size gain for common images.

### DIFF
--- a/public/main.js
+++ b/public/main.js
@@ -1,0 +1,83 @@
+(function() {
+	'use strict';
+
+	// Selectors used in the comparison component
+	var selectors = {
+		component: '[data-component=comparison]',
+		roles: {
+			v1Image: '[data-component-role=v1-image]',
+			v1ImageFormat: '[data-component-role=v1-image-format]',
+			v1ImageSize: '[data-component-role=v1-image-size]',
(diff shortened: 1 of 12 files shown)


public/main.js
```diff
@@ -0,0 +1,88 @@
...
+	});
+
+	// Create a comparison component
+	function createComparison(element) {
+
+		// Get all the component elements
+		var elements = {
+			main: element
+		};
+		Object.keys(selectors.roles).map(function(key) {
```
JakeChampion: As we are not storing the array returned via `map` I think this should be a `forEach` instead.

scripts/fetch-comparison-data.js
```diff
@@ -0,0 +1,33 @@
+#!/usr/bin/env node
+'use strict';
+
+const comparisonImagesPath = `${__dirname}/../data/comparison-images.json`;
```
rowanmanning: Shouldn't be needed, as we're not passing this string to a Windows native command-line utility http://stackoverflow.com/a/9799490

Issue #73, opened 2016-10-06
Change compression level when dpr is applied


rowanbeentje:
I discussed this last week with Jake but worth explicitly raising here; after comparing a few images side-by-side with the v1 images, noticed that the new service was producing images that were much larger.  Looking more closely, a substantial part of this seems to be only when the `dpr` value is set - the old service substantially increased the compression level when dpr was applied.

This is hinted at in the old documentation (not yet ported?):

>Device Pixel Ratio, i.e. number of image pixels per CSS pixel. The default is 1, which gives the standard 1:1 ratio of image pixels to CSS pixels. Set to 2 for high-DPI ("Retina") displays. Request to the image service with width=100&dpr=2 will output image 200 pixels wide, but optimized to be displayed with CSS width:100px
>Note that browsers don't automatically adjust intrinsic size of images with higher DPI, so you need'll need to use some "responsive image" solution, e.g. set explicit width or height, use image-set() or srcset attribute.

Comparing images from v1, using the additional DPR images Jake added last week:
https://image.webservices.ft.com/v1/images/raw/ftcms:2dac88f0-b273-4e48-9c7d-9078db39db55?source=app&width=1000&fit=scale-down - 1000px wide, dpr 1 - **144k**
https://image.webservices.ft.com/v1/images/raw/ftcms:2dac88f0-b273-4e48-9c7d-9078db39db55?source=app&width=500&fit=scale-down&dpr=2 - 1000px wide, dpr2 - **96k**

From v2:
https://www.ft.com/origami/service/image/v2/images/raw/ftcms:2dac88f0-b273-4e48-9c7d-9078db39db55?source=app&width=1000&fit=scale-down - 1000px wide, dpr 1 - **192k**
https://www.ft.com/origami/service/image/v2/images/raw/ftcms:2dac88f0-b273-4e48-9c7d-9078db39db55?source=app&width=500&fit=scale-down&dpr=2 - 1000px wide, dpr 2 - **192k**

(Do the visual compression levels approximately match otherwise?  Significantly larger images anyway, but I'll raise a separate issue for that if necessary)

rowanmanning:
The work in #91 gets us some of the way towards solving this. We may be able to tweak compression levels a bit more in a later PR.

Using the image `ftcms:2dac88f0-b273-4e48-9c7d-9078db39db55` and the transform `width=500&fit=scale-down&dpr=2`:

<dl>
<dt>Image Service V1</dt>
<dd>~94k</dd>
[...]

PR #91, opened 2016-10-19
Tweak the quality values when image DPR is set

rowanmanning:
We now have a different quality map for when a DPR is set, this ensures
that high-DPR images aren't huge in terms of file size.

This should address #73

(yes I know there's still an imgix transformer in there, I'll remove in another PR)

### DIFF
--- a/data/comparison-images.json
+++ b/data/comparison-images.json
@@ -148,7 +148,61 @@
 		"uri": "ftcms:3520df36-8c20-11e6-8cb7-e7ada1d123b1",
 		"transform": "width=250&quality=lossless",
 		"v1ImageFormat": "image/png",
-		"v1ImageSize": "87296"
+		"v1ImageSize": "79304"
+	},
+	{
+		"name": "Lowest and DPR 2",
+		"description": "An image with the quality transform set to 'lowest' and a DPR of 2.",
+		"shouldMatch": true,
+		"uri": "ftcms:03b32b22-91a0-11e6-a72e-b428cb934b78",
+		"transform": "width=250&quality=lowest&dpr=2",
+		"v1ImageFormat": "image/jpeg",
+		"v1ImageSize": "6011"
+	},
+	{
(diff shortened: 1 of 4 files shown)


data/comparison-images.json
```diff
@@ -237,7 +291,7 @@
 		"uri": "fticon-v1:cross",
 		"transform": "width=100&format=png",
 		"v1ImageFormat": "image/png",
-		"v1ImageSize": "719"
```
JakeChampion: because of the newly released fticons?

lib/image-transform.js
```diff
@@ -150,7 +150,11 @@ module.exports = class ImageTransform {
 	setQuality(value = 'medium') {
 		const errorMessage = `Image quality must be one of ${ImageTransform.validQualities.join(', ')}`;
 		const quality = ImageTransform.sanitizeEnumerableValue(value, ImageTransform.validQualities, errorMessage);
-		this.quality = ImageTransform.qualityValueMap[quality];
+		if (this.dpr && this.dpr > 1) {
+			this.quality = ImageTransform.qualityValueMapDpr[quality];
```
JakeChampion: 👍

lib/image-transform.js
```diff
@@ -461,8 +465,8 @@ module.exports = class ImageTransform {
 	 */
 	static get qualityValueMap() {
 		return {
-			lowest: 35,
-			low: 55,
+			lowest: 30,
```
JakeChampion: Why did these also change?

test/unit/lib/image-transform.js
```diff
@@ -206,6 +206,20 @@ describe('lib/image-transform', () => {
 				assert.strictEqual(instance.getQuality(), 100);
 			});
 
+			describe('when `value` is valid and the `dpr` property is set', () => {
```
rowanmanning: Done

PR #96, opened 2016-10-20
Use PREFERRED_HOSTNAME as hostname if it is set

JakeChampion:
The way to get svg-tinting to work from a local running image-service is to expose it to the internet using something like ngrok.io. In order for this to work we need to override what the hostname is.

### DIFF
--- a/config.js
+++ b/config.js
@@ -5,7 +5,7 @@ module.exports = {
 	cloudinaryAccountName: process.env.CLOUDINARY_ACCOUNT_NAME,
 	customSchemeStore: process.env.CUSTOM_SCHEME_STORE,
 	environment: process.env.NODE_ENV || 'development',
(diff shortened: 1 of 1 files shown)


rowanmanning:
This still won't work locally, will it? Because Cloudinary can't access your localhost. The issue is that with SVGs they also have to pass through the tinting endpoint. You can fix locally by setting a `HOSTNAME` environment variable to `origami-image-service-qa.herokuapp.com` which offloads the actual tinting onto QA. Might be easier to discuss when you're in-office :)

JakeChampion:
@rowanmanning I've updated this PR to not take the port into account as that would not work. It now takes the PREFERRED_HOSTNAME into account instead.

rowanmanning:
I've updated the docs, `PREFERRED_HOSTNAME` is no longer referenced anywhere

PR #99, opened 2016-10-21
Set the proxy timeout to 2 minutes

rowanmanning:


rowanmanning:
Actually we don't need this – there is no default timeout, it will just proxy forever

PR #100, opened 2016-10-24
Present Cloudinary errors as HTML error pages

rowanmanning:


lib/image-service.js
```diff
@@ -84,9 +84,18 @@ function createProxy(errorHandler) {
...
 	// Handle proxy responses, allowing modification of a response
 	// before sending it to the user
-	function proxyResponseHandler(proxyResponse, request) {
+	function proxyResponseHandler(proxyResponse, request, response) {
 		const originalHeaders = proxyResponse.headers;
 
+		// If we have a Cloudinary error, save it to the response
+		// and don't assume the response is an image
+		if (originalHeaders['x-cld-error']) {
+			proxyResponse.headers = {};
```
JakeChampion: Could you add a comment about why we are resetting the headers

lib/response-interceptor.js
```diff
@@ -0,0 +1,39 @@
...
+
+function createResponseInterceptor(response, options) {
+	const originals = backUpResponseMethods(response);
+	['writeHead', 'write', 'end'].forEach(methodName => {
+		response[methodName] = function() {
+			if (methodName === 'end') {
+				restoreResponseMethods(response, originals);
+			}
+			if (options.condition()) {
+				return (options[methodName] || noop).apply(response, arguments);
```
JakeChampion: When this line uses the `noop` function, I think it may error because I don't think you can use `apply` or `bind` on arrow functions as they have no context (`this`).

lib/routes/v2/images-raw.js
```diff
@@ -1,15 +1,38 @@
...
 
 const convertToCmsScheme = require('../../middleware/convert-to-cms-scheme');
+const createResponseInterceptor = require('../../response-interceptor');
 const getCmsUrl = require('../../middleware/get-cms-url');
 const httpError = require('http-errors');
 const mapCustomScheme = require('../../middleware/map-custom-scheme');
 const processImageRequest = require('../../middleware/process-image-request');
 
+process.on('uncaughtException', error => {
+	console.log(error.stack);
```
JakeChampion: I think this should be `console.error` to ensure it prints to `stderr` instead of `stdout`.

lib/routes/v2/images-raw.js
```diff
@@ -1,15 +1,38 @@
...
+
 module.exports = (app, router) => {
 
 	// Proxy image middleware
-	function proxyImage(request, response) {
+	function proxyImage(request, response, next) {
+
+		// We have to intercept the response so that we can
+		// turn Cloudinary errors into useful HTML error
+		// pages – normally they are returned as GIFs. So if
```
JakeChampion: We should ask Cloudinary why they return errors as GIFs

lib/response-interceptor.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+// NOTE: this is copied largely from Commandeer, and is
+// definitely a candidate for splitting into a new open
```
JakeChampion: 👍

lib/response-interceptor.js
JakeChampion: My understanding was incorrect, arrow functions will ignore the context parameter of `bind`/`call`/`apply`.

PR #102, opened 2016-10-24
Always serve PNGs when quality is lossless

### LINKED ISSUE #101: Lossless images should always be PNG
For compatibility with v1, images that have a `quality` of `lossless` should always be a PNG. Possibly we could introduce a break where WebP is used if available.

rowanmanning:
Resolves #101


JakeChampion:
I'd be really tempted to introduce a break here as well. Are we at our limiting of breaking changes allowed for a major? (Always hesitant to introduce too many breaks as migration becomes harder)

PR #117, opened 2016-10-27
Purge webpages on deployment

JakeChampion:


scripts/purge.js
```diff
@@ -0,0 +1,56 @@
...
+	'/v2',
+	'/v2/docs/api',
+	'/v2/docs/compare',
+	'/v2/docs/migration'
+];
+
+const endpoints = paths.map(path => hostname + path);
+
+if (!DRY_RUN) {
+	const fetch = require('node-fetch');
```
rowanmanning: `node-fetch` isn't included in the repo. [We are including `isomorphic-fetch`](https://github.com/Financial-Times/origami-image-service/blob/master/package.json#L33) though, maybe use that?

PR #131, opened 2016-11-02
Fix memory leaks

rowanmanning:
There are a few memory leaks, all relating to our use of `fetch`. I've replaced it with `request` in all of the places where it can affect memory usage.


JakeChampion:
This might be good to circulate across to other teams which use fetch in Node. /cc @matthew-andrews @richard-still-ft

matthew-andrews:
interesting!

PR #145, opened 2016-11-08
Add the images/metadata endpoint

rowanmanning:
This matches the `images/metadata` endpoint from v1, minus the cache
object. This allows users to check image dimensions before requesting
the actual image data.

### DIFF
--- a/lib/middleware/get-image-meta.js
+++ b/lib/middleware/get-image-meta.js
@@ -1,0 +1,45 @@
+'use strict';
+
+const httpError = require('http-errors');
+const probe = require('probe-image-size');
+
+module.exports = getImageMeta;
+
+function getImageMeta(request, response, next) {
+	// Get the image meta data
+	probe({
+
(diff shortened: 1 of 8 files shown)


lib/middleware/get-image-meta.js
```diff
@@ -0,0 +1,45 @@
...
+
+function getImageMeta(request, response, next) {
+	// Get the image meta data
+	probe({
+
+		// Request the applied transform
+		url: request.appliedTransform,
+
+		// We need to pass on these client headers to support WebP.
+		// Note: JPEG XR is not supported by the library we use to
```
JakeChampion: Perhaps we should open a feature-request issue and see if they are interested in adding the feature themselves and/or happy for us to create a PR with it?

rowanmanning: https://github.com/nodeca/probe-image-size/issues/10

PR #160, opened 2016-11-21
Add Make target for generating whitesource reports

JakeChampion:
WhiteSource is a tool FT have purchased which enables us to track legal and technical issues within our dependencies. We can ensure we are using dependencies with licences we agree with and that we are notified about dependencies with security vulnerabilities in a timely manner.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -14,6 +14,13 @@ verify-coverage:
 	@istanbul check-coverage --statement $(EXPECTED_COVERAGE) --branch $(EXPECTED_COVERAGE) --function $(EXPECTED_COVERAGE)
 	@$(DONE)
 
+whitesource:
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -14,6 +14,13 @@ verify-coverage:
 	@istanbul check-coverage --statement $(EXPECTED_COVERAGE) --branch $(EXPECTED_COVERAGE) --function $(EXPECTED_COVERAGE)
 	@$(DONE)
 
+whitesource:
+ifndef WHITESOURCE_API_KEY
+	$(error WHITESOURCE_API_KEY is not set, change requests cannot be created. You can find the key in LastPass)
```
rowanmanning: "change requests cannot be created"

PR #182, opened 2016-12-01
Simplify routing

rowanmanning:
CDNs can now send an FT-Origami-Service-Base-Path header to change the
base path across the site documentation. This ensures that when the CDN
performs rewrites, they don't break. It also allows us to access the
service from either ft.com/__origami/... or the heroku URLs without
having to mount routes twice.

### DIFF
--- a/lib/middleware/get-base-path.js
+++ b/lib/middleware/get-base-path.js
@@ -1,0 +1,15 @@
+'use strict';
+
+module.exports = getBasePath;
+
+function getBasePath(request, response, next) {
+	let basePath = request.headers['x-service-base-path'] || '/';
+	if (basePath.substr(0, 1) !== '/') {
+		basePath = `/${basePath}`;
+	}
+	if (basePath.substr(-1, 1) !== '/') {
+		basePath = `${basePath}/`;
+	}
+	request.basePath = response.locals.basePath = basePath;
+	next();
+}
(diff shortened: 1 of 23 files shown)


lib/middleware/get-base-path.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = getBasePath;
+
+function getBasePath(request, response, next) {
+	let basePath = request.headers['x-service-base-path'] || '/';
+	if (basePath.substr(0, 1) !== '/') {
+		basePath = `/${basePath}`;
+	}
+	if (basePath.substr(-1, 1) !== '/') {
```
JakeChampion: I think this would be more readable as `!basePath.endsWith('/')` in place of `basePath.substr(-1,1) !== '/'`.

lib/middleware/get-base-path.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+
+module.exports = getBasePath;
+
+function getBasePath(request, response, next) {
+	let basePath = request.headers['x-service-base-path'] || '/';
```
rowanmanning: As discussed in person, nah we shouldn't need to

rowanmanning:
I've made your requested changes, @JakeChampion

PR #187, opened 2016-12-07
Add more info to the release log description

JakeChampion:


JakeChampion:
@rowanmanning I can't yet add this to the prod change request task as that does not run on circle ci. When we move to releasing via github release tags we can add these tot he change request prod task.

PR #194, opened 2017-01-23
Switch from n-express to Origami Service

rowanmanning:
This is ready now. If we merge this then we need to test extensively on QA before promoting to production. Things that need testing in particular:

  - [ ] The meta endpoints (__about, __health, __gtg)
  - [ ] That metrics are still logged to Graphite
  - [ ] That errors are sent to Sentry
  - [ ] That logs are in the correct format when sent to Splunk


lib/health-checks.js
```diff
@@ -4,82 +4,92 @@ const cloudinary = require('cloudinary');
...
+	constructor(options) {
 
 		cloudinary.config({
-			cloud_name: config.cloudinaryAccountName
+			cloud_name: options.cloudinaryAccountName
 		});
-		customSchemeStoreBucket = config.customSchemeStore;
-		customSchemeCheckUrl = (config.customSchemeStore || '').replace(/\/+$/, '') + '/fticon/v1/cross.svg';
-		cloudinaryCheckUrl = cloudinary.url('http://im.ft-static.com/content/images/a60ae24b-b87f-439c-bf1b-6e54946b4cf2.img', {
+		this.cloudinaryCheckUrl = cloudinary.url('http://im.ft-static.com/content/images/a60ae24b-b87f-439c-bf1b-6e54946b4cf2.img', {
```
JakeChampion: Nit: Switch to https instead of http for the check url

lib/health-checks.js
```diff
@@ -4,82 +4,92 @@ const cloudinary = require('cloudinary');
...
+				technicalSummary: 'Transforms an image with Cloudinary and checks that it responds successfully',
+				panicGuide: 'Check the cloudinary status page http://status.cloudinary.com/'
+			}
+		};
 
-		pingServices();
-		setInterval(pingServices, pingInterval);
-	},
+		this.retrieveData();
+		setInterval(this.retrieveData.bind(this), pingInterval);
```
JakeChampion: Question: Should we unreference this interval? https://nodejs.org/api/timers.html#timers_timeout_unref

lib/health-checks.js
```diff
@@ -4,82 +4,92 @@ const cloudinary = require('cloudinary');
...
+	}
+
+	retrieveData() {
+		this.pingService('cloudinary', this.cloudinaryCheckUrl);
+		this.pingService('customSchemeStore', this.customSchemeCheckUrl);
+	}
 
-	// Ping a service and record its status
 	pingService(name, url) {
 		if (!!process.env.TEST_HEALTHCHECK_FAILURE) {
```
JakeChampion: Question: What do you think to this be a property of the `options` object given to the constructor instead of reading from the environment?

lib/health-checks.js
```diff
@@ -4,82 +4,92 @@ const cloudinary = require('cloudinary');
...
-			name: 'Image sets can be retrieved from the custom scheme store',
-			ok: healthChecks.statuses.customSchemeStore,
-			severity: 2,
-			businessImpact: 'Users may not be able to use custom schemes (e.g. fticon) if they\'re not pre-cached',
-			technicalSummary: 'Hits the given url and checks that it responds successfully',
-			panicGuide: `Check that the bucket ${customSchemeStoreBucket} still exists and that the AWS Region is up`
-		})
+	getGoodToGoFunction() {
+		return () => {
+			return this.getGoodToGoPromise();
```
JakeChampion: Question: Why is this a separate public method on the instance?

lib/health-checks.js
```diff
@@ -4,82 +4,92 @@ const cloudinary = require('cloudinary');
...
+		};
 	}
+
+	getPromise() {
+		return Promise.resolve(this.getStatusArray());
```
JakeChampion: If `this.getStatusArray()` throws an error it will not be wrapped in a `Promise` object. If we are wanting to ensure that `getPromise` only returns a `Promise` object we will have to create a resolved Promise first and call `this.getStatusArray()` in the thenable.
E.G. 
```js
getPromise() {
    return Promise.resolve().then(() => this.getStatusArray());
}
```

lib/health-checks.js
```diff
@@ -4,82 +4,92 @@ const cloudinary = require('cloudinary');
...
+	}
+
+	getGoodToGoPromise() {
+		return Promise.resolve(this.getStatusArray().every(status => status.ok));
```
JakeChampion: Same comment as https://github.com/Financial-Times/origami-image-service/pull/194/files?short_path=04c6e90#r97521985

lib/image-service.js
```diff
@@ -1,57 +1,30 @@
...
-	app.proxy = createProxy(errorHandler);
-	app.imageServiceConfig = config;
+	const healthChecks = new HealthChecks(options);
+	options.healthCheck = healthChecks.getFunction();
+	options.goodToGoTest = healthChecks.getGoodToGoFunction();
+	options.about = require('../about.json');
 
-	healthChecks.init(config);
+	const app = origamiService(options);
+	app.enable('case sensitive routing');
```
JakeChampion: Question: Should this be in Origami Service?

lib/routes/v2/docs/url-builder.js
```diff
@@ -1,12 +1,12 @@
...
 
-const oneWeek = 60 * 60 * 24 * 7;
+const cacheControl = require('@financial-times/origami-service').middleware.cacheControl;
 const querystring = require('querystring');
 
 module.exports = app => {
 
 	// v2 url-builder page
-	app.get('/v2/docs/url-builder', (request, response) => {
+	app.get('/v2/docs/url-builder', cacheControl({maxAge: '7d'}), (request, response) => {
```
JakeChampion: Question: If the request errors after the `cacheControl` middleware has executed, would the error response have a `maxAge` of 7 days?

npm-shrinkwrap.json
```diff
@@ -1,2404 +0,0 @@
-{
```
JakeChampion: Looks like we lost the shrinkwrap? Can we add this back in?

lib/health-checks.js
rowanmanning: I'm not sure, we're not doing this in the health-checks for other services so might be worth investigating as a separate piece of work. Jotting down to create issues later.

lib/health-checks.js
rowanmanning: For ease of testing more than anything. Express Webservice expects a function which returns a promise and I thought it'd be nicer to split into smaller functions to generate each part.

lib/health-checks.js
rowanmanning: `getStatusArray` in its current form can't error, it iterates over some object keys and returns a property that's guaranteed to be there (aside from code in another module messing with it)

lib/routes/v2/docs/url-builder.js
rowanmanning: Good question! Currently yes, need to patch Origami Service to remove cache-control directives in the error handler: https://github.com/Financial-Times/origami-service/issues/33

lib/image-service.js
JakeChampion: Ah yes, ergh, I think it is better to have case sensitive routing on by default and have build service be the exception in our services due to backwards compat.

lib/health-checks.js
```diff
@@ -4,82 +4,93 @@ const cloudinary = require('cloudinary');
...
+				panicGuide: `Check that the bucket ${this.customSchemeCheckUrl} still exists and that the AWS Region is up`
+			},
+			cloudinary: {
+				id: 'cloudinary-image-transform',
+				name: 'Images can be transformed with Cloudinary',
+				ok: true,
+				severity: 2,
+				businessImpact: 'Users may not be able to view images if they\'re not pre-cached',
+				technicalSummary: 'Transforms an image with Cloudinary and checks that it responds successfully',
+				panicGuide: 'Check the cloudinary status page http://status.cloudinary.com/'
```
JakeChampion: The url can now be https :-D

PR #207, opened 2017-02-02
Fix the meta-data endpoint for JPEG XR images

rowanmanning:


lib/middleware/get-image-meta.js
```diff
@@ -7,6 +7,13 @@ const probe = require('probe-image-size');
 module.exports = getImageMeta;
 
 function getImageMeta(request, response, next) {
+
+	// Prevent JPEG-XR from causing issues
+	let acceptHeader = request.headers.accept;
+	if (acceptHeader && acceptHeader.indexOf('image/jxr') !== -1) {
```
JakeChampion: As we are trying to find out if the accept header includes `image/jxr`, we should use the `includes` method instead of `indexOf`.

rowanmanning: Good spot, I keep thinking this doesn't work yet but it does

PR #219, opened 2017-03-01
Handle broken Cloudinary responses

rowanmanning:
This PR correctly handles non-200 Cloudinary responses which are so broken that they don't send their usual `X-Cld-Error` header. We saw this happen recently with the AWS failures.


lib/image-service.js
```diff
@@ -94,6 +95,25 @@ function createProxy(errorHandler) {
 			return;
 		}
 
+		// If we have a 40x/50x response at this point, Cloudinary
+		// is so broken that it can't even send its own X-CLD-Error
+		// header. We need to handle this to prevent these errors
+		// from being cached
+		if (proxyResponse.statusCode >= 400) {
+			proxyResponse.headers = {};
+			response.cloudinaryError = httpError(proxyResponse.statusCode);
```
JakeChampion: If this property exists, we throw our own error message - https://github.com/Financial-Times/origami-image-service/blob/master/lib/routes/v2/images.js#L62

PR #229, opened 2017-03-08
Unique surrogate key for each image being transformed ✨

JakeChampion:
If this lands, it would be possible to add another purge endpoint which is used to purge a specific image.

Example:
Image wanting to be purged --
https://www.ft.com/__origami/service/image/v2/images/raw/ftsocial-v2:twitter?source=docs
How to purge the image (You would need the api header set with the correct api key) --
https://www.ft.com/__origami/service/image/v2/images/purge/ftsocial-v2:twitter?source=docs
That purge endpoint could grab the surrogate key which is specific to that original image and purge it.

### DIFF
--- a/lib/image-service.js
+++ b/lib/image-service.js
@@ -141,7 +141,7 @@ function createProxy(errorHandler) {
 
 		const normaliseKeys = key => (key || '').replace(/[^a-z0-9-]/gi, '').toLowerCase();
 
-		proxyResponse.headers['Surrogate-Key'] = `${normaliseKeys(keyForAllImages)} ${normaliseKeys(keyForImageType)} ${normaliseKeys(keyForScheme)}`;
+		proxyResponse.headers['Surrogate-Key'] = `${normaliseKeys(keyForAllImages)} ${normaliseKeys(keyForImageType)} ${normaliseKeys(keyForScheme)} ${request.params.schemeUrl}`;
 
 		if (request.transform && request.transform.getDpr()) {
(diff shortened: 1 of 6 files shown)


lib/image-service.js
```diff
@@ -141,7 +141,7 @@ function createProxy(errorHandler) {
 
 		const normaliseKeys = key => (key || '').replace(/[^a-z0-9-]/gi, '').toLowerCase();
 
-		proxyResponse.headers['Surrogate-Key'] = `${normaliseKeys(keyForAllImages)} ${normaliseKeys(keyForImageType)} ${normaliseKeys(keyForScheme)}`;
+		proxyResponse.headers['Surrogate-Key'] = `${normaliseKeys(keyForAllImages)} ${normaliseKeys(keyForImageType)} ${normaliseKeys(keyForScheme)} ${request.params.schemeUrl}`;
```
JakeChampion: I thought we should normalise this surrogate key because I think it it simpler for purging to keep the string identical to the one being requested.

JakeChampion:
If this lands, it would be possible to add another purge endpoint which is used to purge a specific image.

Example:
Image wanting to be purged -- 
`https://www.ft.com/__origami/service/image/v2/images/raw/ftsocial-v2:twitter?source=docs`
How to purge the image (You would need the api header set with the correct api key) --
`https://www.ft.com/__origami/service/image/v2/images/purge/ftsocial-v2:twitter?source=docs`
That purge endpoint could grab the surrogate key which is specific to that original image and purge it.

PR #236, opened 2017-03-09
Do not cache the purging endpoints

JakeChampion:


lib/middleware/no-cache.js
```diff
@@ -0,0 +1,7 @@
+'use strict';
+
+module.exports = () => (req, res, next) => {
+	res.set('Cache-Control', 'private, max-age=0, no-cache');
+	res.set('Surrogate-Control', 'private, max-age=0, no-cache');
```
rowanmanning: Should `Surrogate-Control` be added to the [Origami Service Cache Control middleware](https://github.com/Financial-Times/origami-service/blob/master/lib/middleware/cache-control.js)? I can imagine it being useful in more places than this. For now this is fine but a suggestion for future updates :)

PR #238, opened 2017-03-10
Add ability to purge URLs from cloudinary

JakeChampion:


lib/purge-from-fastly.js
```diff
@@ -0,0 +1,59 @@
...
+						if (error) {
+							console.log(`Could not purge ${url} from Fastly, will re-attempt in an hour.`);
+							addToPurgeQueue(url);
+						} else {
+							console.log(`Purged ${url} from Fastly`);
+						}
+					});
+				}
+
+				delete urlsToPurge[nowHour];
```
rowanmanning: Do we need to set `timer` to a falsy value here too? Otherwise I think this will only ever call `setTimeout` once

PR #261, opened 2017-03-22
Log Cloudinary usage information to Graphite

rowanmanning:


lib/cloudinary-metrics.js
```diff
@@ -0,0 +1,44 @@
...
+					this.metrics.count('cloudinary.transformations.usage', result.transformations.usage);
+					this.metrics.count('cloudinary.transformations.limit', result.transformations.limit);
+					this.metrics.count('cloudinary.objects.usage', result.objects.usage);
+					this.metrics.count('cloudinary.objects.limit', result.objects.limit);
+					this.metrics.count('cloudinary.bandwidth.usage', result.bandwidth.usage);
+					this.metrics.count('cloudinary.bandwidth.limit', result.bandwidth.limit);
+					this.metrics.count('cloudinary.storage.usage', result.storage.usage);
+					this.metrics.count('cloudinary.storage.limit', result.storage.limit);
+				})
+				.catch(error => {
```
JakeChampion: If it throws, should we clear the interval?

rowanmanning: I thought about it but then that could prevent us from getting any data for up to 24 hours. I'd rather risk an error getting logged every 5 mins and let it fix itself

PR #271, opened 2017-04-10
Add a healthcheck for UPP images

JakeChampion:
This is waiting on a non-cached endpoint that we should be able to poll against.

### DIFF
--- a/lib/health-checks.js
+++ b/lib/health-checks.js
@@ -17,7 +17,7 @@ module.exports = class HealthChecks {
 		const customSchemeCheckBaseUrl = options.customSchemeStore.replace(/\/+$/, '');
 		this.customSchemeCheckUrl = `${customSchemeCheckBaseUrl}/fticon/v1/cross.svg`;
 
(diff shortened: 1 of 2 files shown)


lib/health-checks.js
```diff
@@ -35,6 +37,15 @@ module.exports = class HealthChecks {
...
 				panicGuide: 'Check the cloudinary status page https://status.cloudinary.com/'
+			},
+			uppImages: {
+				id: 'upp-images',
+				name: 'Images can be retrieved from the Universal Publishing Platform',
+				ok: true,
+				severity: 2,
+				businessImpact: 'Users may not be able to view images served by Universal Publishing Platform',
+				technicalSummary: 'Hits the given url and checks that it responds successfully',
+				panicGuide: `Check that the UPP endpoint ${this.customSchemeCheckUrl} still exists and that the AWS Region is up`
```
rowanmanning: Should be `this.uppImagesUrl`

PR #282, opened 2017-08-23
Enable purging to work from ft.com and improve the documentation

JakeChampion:
# What's changed?

- Revert the surrogate keys back to their original form. Updating a surrogate key value is a complicated task to complete for so many requests. If we updated the surrogate keys and then decided to purge one of the new surrogate keys (E.G. All webp images), it would only purge the requests that were sent with the new surrogate key and not the old one (E.G. Not all webp images would be purged).

- Updated the purging documentation to remove the incorrect surrogate-key examples and instead to inform the reader which surrogate key corresponds to the specific task they are trying to achieve.

- Made the `/v2/purge/` route use the origami-base-path value to enable this route to work correctly on FT.com

- Added the purging documentation paths to the in-built purging mechanism which is used when deploying the application to QA and production environments. This means that the documentation pages won't ever show old versions of the documentation.

### DIFF
--- a/lib/image-service.js
+++ b/lib/image-service.js
@@ -169,7 +169,7 @@ function createProxy(errorHandler) {
 		const normaliseKeys = (key = '') => key.toLowerCase();
 		const encodeKeys = (key = '') => base64.encode(utf8.encode(key));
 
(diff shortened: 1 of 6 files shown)


alicebartlett:
Good catch on the surrogate keys change.
This LGTM but I think a second opinion might be good as I don't know a lot about caching.

PR #286, opened 2017-09-12
Add immutable cache cloudinary flag for FTCMS images

JakeChampion:
I looked at adding the immutable cache control setting into image service for cloudinary requests for FT CMS images (These are the only images we know 100% are immutable). -- http://cloudinary.com/product_updates/immutable_cache_control_response


JakeChampion:
@rowanmanning Let's merge and release tomorrow morning

PR #290, opened 2017-09-22
Enable Node 8 on Heroku and CI

JakeChampion:
Node 8 does not accept null to assign a random port. Node 6/7/8 all accept 0 to assign a random port.

>Omit the port argument, or use a port value of 0, to have the operating system assign a random port, which can be retrieved by using server.address().port after the 'listening' event has been emitted. [source](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_server_listen_port_hostname_backlog_callback)

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "url": "https://github.com/Financial-Times/origami-image-service.git"
   },
   "engines": {
-    "node": "^6.11.1",
-    "npm": "^4"
+    "node": ">6.11.1",
+    "npm": ">4"
   },
   "main": "./lib/image-service.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -16,8 +16,8 @@
     "url": "https://github.com/Financial-Times/origami-image-service.git"
   },
   "engines": {
-    "node": ">6.11.1",
-    "npm": ">4"
+    "node": "6.11.1 - 8",
+    "npm": "4 - 5"
   },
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -16,8 +16,8 @@
     "url": "https://github.com/Financial-Times/origami-image-service.git"
   },
   "engines": {
-    "node": "^6.11.1",
-    "npm": "^4"
+    "node": ">6.11.1",
+    "npm": ">4"
```
rowanmanning: Should we maybe pin these at a major so that it doesn't automatically start installing Node.js 9.x when it's released? I think I'd rather stick with stable/LTS releases of Node.js.

test/integration/setup.js
```diff
@@ -17,7 +17,7 @@ before(function() {
 		defaultLayout: 'main',
 		environment: 'test',
 		log: mockLog,
-		port: null,
+		port: 0,
```
rowanmanning: Does this work? I've never seen a `0` port used before, I've used `undefined` with Node.js 8 I think.

JakeChampion: This works :-).

>Omit the port argument, or use a port value of 0, to have the operating system assign a random port, which can be retrieved by using server.address().port after the 'listening' event has been emitted. [source](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_server_listen_port_hostname_backlog_callback)

Issue #173, opened 2016-11-24
Write integration tests to ensure that WebP/JPEG XR varying works


rowanmanning:
We should add some integration tests which confirm that the app always sends the correct `Content-Type` and `FT-Image-Format` headers when the `format` parameter is `auto`. We'd need to cover the following cases:

| Req `Accept` | Req `User-Agent` | Res `Content-Type`   | Res `FT-Image-Format` |
|--------------|------------------|----------------------|-----------------------|
| `*/*`        | -                | `image/jpeg`         | `default`             |
| `image/webp` | -                | `image/webp`         | `webp`                |
| `image/jxr`  | -                | `image/vnd.ms-photo` | `jpegxr`              |
| `*/*`        | Chrome           | `image/jpeg`         | `default`             |
| `image/webp` | Chrome           | `image/webp`         | `webp`                |
| `image/jxr`  | Chrome           | `image/vnd.ms-photo` | `jpegxr`              |
| `*/*`        | IE               | `image/jpeg`         | `default`             |
| `image/webp` | IE               | `image/webp`         | `webp`                |
| `image/jxr`  | IE               | `image/vnd.ms-photo` | `jpegxr`              |

PR #308, opened 2017-11-09
Add tests for jpeg output format when 'format' parameter is auto

Limess:
Used the existing defined `.jpg` from S3 for lack of knowledge of alternatives.

Closes https://github.com/Financial-Times/origami-image-service/issues/173.


Limess:
The upp image has no extension which (seemingly) results in the content-type being returned as image/jpeg rather than image/vnd.ms-photo for any user agent.

PR #321, opened 2018-01-23
Updating the regex for the new upp image buckets.

peteclark-ft:


lib/middleware/convert-to-cms-scheme.js
```diff
@@ -3,7 +3,7 @@
 module.exports = convertToCmsScheme;
 
 function convertToCmsScheme() {
-	const cmsRegExp = /^(https?:)?\/*(?:prod-upp-image-read\.ft\.com|com\.ft\.imagepublish\.prod(-us)?\.s3\.amazonaws\.com|im\.ft-static\.com\/content\/images)\/([0-9a-f-]+)(?:\.(img|png|jpg|jpeg|gif))?(\?.+)?$/i;
+	const cmsRegExp = /^(https?:)?\/*(?:prod-upp-image-read\.ft\.com|com\.ft\.imagepublish\.(?:upp-)?prod(-us|-eu)?\.s3\.amazonaws\.com|im\.ft-static\.com\/content\/images)\/([0-9a-f-]+)(?:\.(img|png|jpg|jpeg|gif))?(\?.+)?$/i;
```
JakeChampion: Are we okay with this also allowing through `http://com.ft.imagepublish.prod-eu.s3.amazonaws.com/26a7951a-c3e4-11e7-b30e-a7c1c7c13aab`?

peteclark-ft: Good point - no probably not 🤔 .

PR #350, opened 2019-03-29
update packages to latest versions and use latest nodejs version

JakeChampion:


lib/routes/v2/images.js
```diff
@@ -94,7 +94,6 @@ module.exports = app => {
 	// Request with no scheme at all
 	app.get(
 		/^\/v2\/images\/(raw|debug|metadata|purge)\/$/i, function() {
-			console.log('hi');
```
notlee: 😄

test/integration/setup.test.js
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-const imageService = require('../..');
+  const imageService = require('../..');
```
notlee:
```suggestion
const imageService = require('../..');
```

test/unit/lib/middleware/map-custom-scheme.test.js
```diff
@@ -123,7 +123,7 @@ describe('lib/middleware/map-custom-scheme', () => {
 			describe('when the resolved URL does not have a valid extension', () => {
 
 				beforeEach(() => {
-					next.reset();
+					next. resetHistory();
```
notlee: Wow, this space doesn't matter :mind-blown:

notlee:
Oh! The README needs updating if this no longer runs in node 6

PR #351, opened 2019-04-01
Use Node.js 10

rowanmanning:
Our general policy is to only use LTS releases in production, I know we've discussed in-person but should we maybe write it down somewhere? I'm reminded because of this tweet:

> Quick PSA...
> Please only use LTS releases of @nodejs in production.
> Once we end of life a release, stop using it!
> – https://twitter.com/MylesBorins/status/1110960442292256772

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "url": "https://github.com/Financial-Times/origami-image-service.git"
   },
   "engines": {
-    "node": "^11",
+    "node": "^10",
     "npm": "^6"
   },
   "main": "./lib/image-service.js",


chee:
i think this tweet means use LTS at a minimum, as in "do not use end-of-life node", we could very comfortably use `node@latest` in production.

it's the latest stable release. leading edge, not bleeding edge.

rowanmanning:
Responses to this tweet would suggest otherwise https://twitter.com/Andersos/status/1110967197357228032

chee:
@rowanmanning that's very interesting

PR #352, opened 2019-04-10
Image sets

JakeChampion:
Problem:
Origami Image Service currently has hardcoded where the location of Origami ImageSets are, this makes it very difficult to update where an image will be stored because we would have to coordinate between both systems (Image Service and all ImageSets).

Solution:
Let's make each ImageSet a dependency of the Image Service and get the location of the images via `require`ing the ImageSet and using the ImageSet manifest (imageset.json) as that is structured data which follows a specification. This allows each image in each ImageSet to have a completely different URL that doesn't need to follow any pattern whatsoever.

Benefit:
This allows each image in an ImageSet to be stored at a "random" location, such as having the location (URL) contain the SHA-256 hash of the image, thereby making the need for cachebusting an image obsolete as whenever an image is updated it would have a new location.


package.json
```diff
@@ -25,9 +25,17 @@
...
+    "@financial-times/headshot-images": "^1.143.0",
     "@financial-times/health-check": "^1.9.0",
+    "@financial-times/logo-images": "^1.9.2",
+    "@financial-times/origami-brand-images": "^1.1.5",
+    "@financial-times/origami-flag-images": "^1.0.3",
     "@financial-times/origami-service": "^4.0.0",
     "@financial-times/origami-service-makefile": "^6.4.0",
+    "@financial-times/origami-specialist-title-logos": "^1.3.1",
+    "@financial-times/podcast-logos": "^1.1.4",
+    "@financial-times/social-images": "^2.2.2",
```
rowanmanning: Does this mean that we need to bump dependencies in the Image Service every time we release an image set? Would be nice to automate that part.

JakeChampion: Yes, we would look to automate this part, which is why I've added integrations tests for every image in every imageset to ensure that they all work when an imageset is updated. 

We would have to do it manually when an imageset has a new major version though, because that means it has some kind of breaking change for us to deal with.

test/unit/lib/image-transform.test.js
```diff
@@ -976,42 +1029,20 @@ describe('lib/image-transform', () => {
 
 		});
 
-		describe('when `uri` has a path', () => {
```
JakeChampion: @rowanmanning I removed these because I think they were for a feature that was not meant to exist and that I think does not get used.

Do you know if this feature was meant to exist and if it is being used at all?

I'm thinking of searching the logs for this, but it would be a very difficult thing to search for I imagine.

rowanmanning: It was originally added to cater for `next-assets`, because their src directory wasn't flat. However the repo has since been archived and I don't think it's used anywhere. In fact I have to link to an old commit to even demonstrate the non-nested structure https://github.com/Financial-Times/next-assets/tree/4c3a5cb5096a85648b27d74fd2ce2a3065d2bdbf

PR #372, opened 2019-04-23
Increase cache hit ratio by removing parts of cache key which are not needed

JakeChampion:
don't include the host header or the source query parameter in the cache key to increase the cache hit ratio (host can be 3 different values, source can be any value :| )

header and source values are not used in the generation of an image and so can be removed from the cache key. 

We store in the cache key whether a source parameter did exist because we serve an error page if the source parameter is missing.

This should make FTCMS images have a much better cache hit ratio because they are requested with source parameters of `next`, `app`, `fruitcake` etc.

### DIFF
--- a/fastly/vcl/image-service.vcl
+++ b/fastly/vcl/image-service.vcl
@@ -1,3 +1,5 @@
+import querystring;
+
 sub set_backend {
 	# The Fastly macro is inserted before the backend is selected because the
 	# macro has the code which defines the values avaiable for req.http.Host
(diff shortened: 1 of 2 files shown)


fastly/vcl/image-service.vcl
```diff
@@ -109,8 +111,17 @@ sub vcl_hash {
 		call breadcrumb_hash;
 	}
 
-	set req.hash += req.http.host;
-	set req.hash += req.url;
+	# Do not add the host header because it can have 1 of 3 values: origami-image-service.in.ft.com, origami-image-service-eu.herokuapp.com , origami-image-service-us.herokuapp.com
+	# set req.hash += req.http.host;
+	
+	# Do include the source query parameter in the url when adding the url to the hash but do include whether the url did have a source parameter at all
```
notlee:
```suggestion
	# Do not include the source query parameter in the url when adding the url to the hash but do include whether the url did have a source parameter at all
```

PR #406, opened 2019-07-05
Use Content API to locate the url of an image if it exists in the Content API otherwise fallback to original logic

JakeChampion:
This ensures we still work for old Content API v1 images and work for the images coming from the new CMS called Spark

### DIFF
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -43,7 +43,7 @@ function getCmsUrl(config) {
 						return response.data.binaryUrl;
 					}
 				}
-				// First try fetching the v2 image
+				// Second try fetching from FT CMS v2
 				return requestPromise({
 						uri: v2Uri,
 						method: 'HEAD',
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -1,11 +1,12 @@
 'use strict';
 
 const httpError = require('http-errors');
+const axios = require('axios');
 const requestPromise = require('../request-promise');
 
 module.exports = getCmsUrl;
 
-function getCmsUrl() {
+function getCmsUrl(config) {
 	return (request, response, next) => {
 
 		if (!request.params.imageUrl.startsWith('ftcms')) {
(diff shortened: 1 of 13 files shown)


JakeChampion:
Needs some tests.

lib/middleware/get-cms-url.js
```diff
@@ -20,41 +21,58 @@ function getCmsUrl() {
...
+			})
+			.then(response => {
+				if (response.status <= 400) {
+					if (response.data.binaryUrl.length > 0) {
+						return response.data.binaryUrl;
+					}
 				}
-				// If the v2 image can't be found, try v1
-				lastRequestedUri = v1Uri;
+				// First try fetching the v2 image
```
notlee:
```suggestion
				// Second try fetching from FT CMS v2
```

lib/middleware/get-cms-url.js
```diff
@@ -20,41 +21,58 @@ function getCmsUrl() {
...
-			uri: v2Uri,
-			method: 'HEAD',
-			timeout: 10000 // 10 seconds
-		})
-			.then(firstResponse => {
-				// Cool, we've got an image from v2
-				if (firstResponse.statusCode <= 400) {
-					cmsVersionUsed = 'v2';
-					return v2Uri;
+		axios.get(capi, {
```
notlee:
```suggestion
		// First try fetching from Content API
		axios.get(capi, {
```

lib/routes/v2/images.js
```diff
@@ -119,7 +119,7 @@ module.exports = app => {
 	// /v2/images/raw/https://im.ft-static.com/...
 	// /v2/images/debug/http://im.ft-static.com/...
 	app.get(
-		/^\/v2\/images\/(raw|debug|metadata|purge)\/((https?(:|%3A))?(\/|%2F)*(prod-upp-image-read\.ft\.com|com\.ft\.imagepublish|im\.ft-static\.com).+)$/i,
+		/^\/v2\/images\/(raw|debug|metadata|purge)\/((https?(:|%3A))?(\/|%2F)*(prod-upp-image-read\.ft\.com|com\.ft\.imagepublish|im\.ft-static\.com|d1e00ek4ebabms\.cloudfront\.net|cct-images\.ft\.com).+)$/i,
```
notlee: The `convertToCmsScheme` function matches on the production route only `(?:d1e00ek4ebabms\.cloudfront\.net\/production|cct-images\.ft\.com\/production)`, should that be the same here?

test/integration/v2/images-debug.test.js
```diff
@@ -6,29 +6,127 @@ const itRespondsWithStatus = require('../helpers/it-responds-with-status');
...
+			requestedUrl: 'http://d1e00ek4ebabms.cloudfront.net/production/817dd37c-b808-4b32-9db2-d50bdd92372b.jpg',
+			resolvedUrl: 'https://d1e00ek4ebabms.cloudfront.net/production/817dd37c-b808-4b32-9db2-d50bdd92372b.jpg',
+		},
+		httpssparkcustomdomain: {
+			requestedUrl: 'https://cct-images.ft.com/production/817dd37c-b808-4b32-9db2-d50bdd92372b.jpg',
+			resolvedUrl: 'https://d1e00ek4ebabms.cloudfront.net/production/817dd37c-b808-4b32-9db2-d50bdd92372b.jpg',
+		},
+		httpsparkcustomdomain: {
+			requestedUrl: 'http://cct-images.ft.com/production/817dd37c-b808-4b32-9db2-d50bdd92372b.jpg',
+			resolvedUrl: 'https://d1e00ek4ebabms.cloudfront.net/production/817dd37c-b808-4b32-9db2-d50bdd92372b.jpg',
```
notlee: As above, what happens to a non `/production` image?

lib/routes/v2/images.js
JakeChampion:
```suggestion
		/^\/v2\/images\/(raw|debug|metadata|purge)\/((https?(:|%3A))?(\/|%2F)*(prod-upp-image-read\.ft\.com|com\.ft\.imagepublish|im\.ft-static\.com|d1e00ek4ebabms\.cloudfront\.net\/production|cct-images\.ft\.com\/production).+)$/i,
```

test/integration/v2/images-debug.test.js
JakeChampion: They wouldn't be images coming from Spark CMS, so they would not get converted to the ftcms custom scheme and would be treated as regular http/s image requests, which have a lower cache-control set and have a different set of surrogate-keys set.

PR #417, opened 2019-07-16
use new change management api orb

JakeChampion:


.circleci/config.yml
```diff
@@ -23,7 +28,11 @@ jobs:
       - run: terraform apply --auto-approve -var name=origami-image-service-dev.in.ft.com -var domain=origami-image-service-dev.in.ft.com fastly/terraform/
       - run: sleep 60
       - run: "curl -X POST -H \"Fastly-Key: $FASTLY_API_KEY\" https://api.fastly.com/service/$FASTLY_SERVICE_ID_DEV/purge_all"
-      - run: npx -p @financial-times/release-log@^1 release-log --api-key $RELEASE_LOG_API_KEY --service 'Origami Image Service V2' --owner-email 'rowan.manning@ft.com' --notify-channel "origami-deploys" --description 'Release triggered by CI' --environment Development --summary 'Releasing Origami Image Service to Development'
+      - change-api/release-log:
+                  changeApiKey: '${CHANGE_API_KEY}'
```
nl-ria: If you add `CHANGE_API_KEY` as a CircleCI project env variable, you dont have to explicitly set it here. The orb will default to the variable :) One less line of change

.circleci/config.yml
```diff
@@ -1,4 +1,9 @@
-version: 2
+version: 2.1
+
+orbs:
+    change-api: financial-times/change-api
```
Limess: I think orb versioning is mandatory?

.circleci/config.yml
```diff
@@ -23,7 +28,10 @@ jobs:
       - run: terraform apply --auto-approve -var name=origami-image-service-dev.in.ft.com -var domain=origami-image-service-dev.in.ft.com fastly/terraform/
       - run: sleep 60
       - run: "curl -X POST -H \"Fastly-Key: $FASTLY_API_KEY\" https://api.fastly.com/service/$FASTLY_SERVICE_ID_DEV/purge_all"
-      - run: npx -p @financial-times/release-log@^1 release-log --api-key $RELEASE_LOG_API_KEY --service 'Origami Image Service V2' --owner-email 'rowan.manning@ft.com' --notify-channel "origami-deploys" --description 'Release triggered by CI' --environment Development --summary 'Releasing Origami Image Service to Development'
+      - change-api/release-log:
```
Limess: You need `curl` and `jq` to be installed if running the `job` version. The `workflow` version is generally more convenient but this is fine if you prefer it.

JakeChampion: Do you have an example?

Limess: Not entirely sure which part you're asking for an example for, so:

For using the exposed jobs rather than steps:
https://github.com/Financial-Times/prometheus/blob/master/.circleci/config.yml#L305-L316

For installing jq using orbs: https://github.com/CircleCI-Public/jq-orb/blob/master/src/examples/install.yml#L7-L16

nl-ria: Example of using it as a job (in a workflow) - https://github.com/Financial-Times/change-api/blob/master/.circleci/config.yml#L139

Example of using it as command - 

```
orbs:
  jq: circleci/jq@1.9.0
...
...
jobs:	
## must be run in an image with curl installed 
  change-api-release:	
    steps:	
      - checkout	
      - jq/install	
      - change-api/generate:	
          systemCode: "change-request-api" # Required .The system being released, should exist in Biz-Ops.	
          environment: "test" # Required. The environment in which the system is being changed.	
          slackChannels: "ft-changes-test,bot-playground"	
          extraProperties: '{"createdBy":"Change API Orb","MoreInformation":"See here https://github.com/Financial-Times/change-api"}'```

PR #559, opened 2020-06-09
If Content API does not have the image, try to use the original image url and see if it exists

JakeChampion:
When working with unpublished content, the content API will not respond with the image but the original image url does in fact exist and work.

This change will make the image service attempt to use the original image url if it responds with a successful HTTP status code.

This change will make origami image service work for unpublished content being authored within Spark CMS.

### DIFF
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -68,29 +68,28 @@ function getCmsUrl(config) {
 								cmsVersionUsed = 'v1';
 								return v1Uri;
 							}
-							if (request.params.originalImageUrl) {
-								return requestPromise({
-									uri: request.params.originalImageUrl,
-									method: 'HEAD',
-									timeout: 10000 // 10 seconds
-								}).then(thirdResponse => {
-									// Cool, we've got an image from the original image url, which means the image has not yet been published to the Content API.
-									// This usually means the image is being used in an unpublished entry in the Spark CMS.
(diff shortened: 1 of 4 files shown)


lib/middleware/get-cms-url.js
```diff
@@ -68,10 +68,29 @@ function getCmsUrl(config) {
...
+										cmsVersionUsed = 'n/a';
+										return request.params.originalImageUrl;
+									}
+									// If the v1 image can't be found, we error
+									const error = httpError(404, `Unable to get image ${cmsId} from Content API v1 or v2`);
+									error.cacheMaxAge = '30s';
+									throw error;
+								});
+							} else {
+								// If the v1 image can't be found, we error
```
notlee, line 89: maybe we could invert this `if` to error early? would be nice to reduce the indenting a little

JakeChampion: Done in [https://github.com/Financial-Times/origami-image-service/pull/559/commits/8fbc34196840c2687ed9ea5abb99ae19e82deb59 ](https://github.com/Financial-Times/origami-image-service/pull/559/commits/8fbc34196840c2687ed9ea5abb99ae19e82deb59) :-)

origamiserviceuser:
:tada: This PR is included in version v431.1.0 :tada:

taktran:
Awesome, thanks! ❤️ from Spark ✨

PR #615, opened 2020-08-28
use cloudinary's v2 api

JakeChampion:


package.json
```diff
@@ -40,7 +40,7 @@
     "@financial-times/source-param-middleware": "2.0.0",
     "axios": "0.20.0",
     "base-64": "0.1.0",
-    "cloudinary": "1.22.0",
+    "cloudinary": "^1.22.0",
```
notlee, line 43: should we set these without a range and let dependabot do the rest?

notlee:
Is there some cloudinary documentation which shows changes from v1 to v2? It looks like the signiture for some of these methods has changed but this library is maybe handling the v1 signiture?

package.json
JakeChampion: done 👍

JakeChampion:
I tried looking for a migration guide or anything and I found nothing about the v2 api so I'm going to close this because the v1 api looks like it is here to stay.

PR #620, opened 2020-09-02
store images in Cloudinary using content hash as their ID

JakeChampion:
This changes the Origami Image Service's use of Cloudinary's fetch API with their upload API and changes how the images are stored.

Previously we used the image's url as the public ID when storing on Cloudinary. This is now causing issues because the publishing platforms which use the Origami Image Service have begun to use new image urls for the same image, meaning we have thousands upon thousands of duplicate images being stored in Cloudinary. _In a short manual scan over a few minutes I found over four thousand duplicate images._

This commit makes the Origami Image Service download the image first and calculate the SHA256 of the image, using the SHA value as the public ID. This solves the issue of duplicate images because the exact same image served at different URLs will now resolve to the same public ID in Cloudinary.

To set public ID's in Cloudinary is not possible using the fetch API which is why this commit also switches to the upload API.

This change does mean that the Origami Image Service is now downloading the full image before sending it to Cloudinary, which means it will have a slower response time and larger memory usage, this is only an issue for the first time an image is requested though because of all the caching in the Fastly CDN and Cloudinary's CDN, as well as the in-memory cache in the application code.

### DIFF
--- a/lib/middleware/process-image-request.js
+++ b/lib/middleware/process-image-request.js
@@ -1,12 +1,25 @@
 'use strict';
 
+const httpError = require('http-errors');
 const ImageTransform = require('../image-transform');
 const cloudinaryTransform = require('../transformers/cloudinary');
 const url = require('url');
+const axios = require('axios').default;
+const crypto = require('crypto');
+const cloudinary = require('cloudinary').v2;
+const promisify = require('util').promisify;
 
 module.exports = processImage;
 
 function processImage(config) {
+	const cache = Object.create(null);
+	cloudinary.config({
+		cloud_name: config.cloudinaryAccountName,
+		api_key: config.cloudinaryApiKey,
+		api_secret: config.cloudinaryApiSecret
+	});
+	const resource = promisify(cloudinary.api.resource.bind(cloudinary.api));
+	const upload = promisify(cloudinary.uploader.upload.bind(cloudinary.uploader));
 	return (request, response, next) => {
 		let transform;
 		// Add the URI from the path to the query so we can
@@ -34,7 +47,7 @@ function processImage(config) {
 			const tint = (transform.tint ? transform.tint[0] : '');
 			const hostname = (config.hostname || request.hostname);
 			const protocol = (hostname === 'localhost' ? 'http' : 'https');
-			// TODO change svgtint to just svg now that it does more
+			// TODO change svgtint to just svg now that it also sanitizes svgs.
 			transform.setUri(`${protocol}://${hostname}/v2/images/svgtint/${encodedUri}${hasQueryString ? '&' : '?'}color=${tint}`);
 			// Clear the tint so that SVGs converted to rasterised
 			// formats don't get double-tinted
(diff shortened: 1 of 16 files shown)


test/integration/setup.test.js
```diff
@@ -14,7 +14,10 @@ before(function() {
 	return imageService({
 		contentApiKey: process.env.CONTENT_API_KEY,
 		cloudinaryAccountName: 'financial-times', // TODO set up a test account for this?
-		customSchemeStore: process.env.CUSTOM_SCHEME_STORE || 'https://origami-images.ft.com',
+		cloudinaryApiKey: process.env.CLOUDINARY_API_KEY,
+		cloudinaryApiSecret: process.env.CLOUDINARY_API_SECRET,
+		customSchemeCacheBust: process.env.CUSTOM_SCHEME_CACHE_BUST || '',
+		customSchemeStore: process.env.CUSTOM_SCHEME_STORE,
```
JakeChampion, line 20:
```suggestion
		customSchemeStore: process.env.CUSTOM_SCHEME_STORE || 'https://origami-images.ft.com',
```

JakeChampion:
@Financial-Times/origami-core This is ready to review now.

If this change gets approved and merged then the roll-out plan would be this:
- Write a migration script which downloads all the images from Cloudinary and then re-uploads them with their public ID set to the SHA256 of the image.
- Merge this pull-request and deploy to QA.
- Run the migration script.
- Set up Charles Proxy to rewrite production Origami Image Service URLs into ones which go through QA and confirm the images are working correctly by visiting ft.com and other users of the Image Service.
- Release to production.
- After a week in production we delete all images in Cloudinary which were uploaded more than a week ago (before the change was released to production).

notlee:
nice! I'll take a look at this tomorrow :)

lib/middleware/process-image-request.js
```diff
@@ -47,17 +60,75 @@ function processImage(config) {
...
+				if (imageResponse.status >= 400) {
+					throw httpError(imageResponse.status);
+				}
+				if (imageResponse.headers['content-type'].includes('text/html')) {
+					throw httpError(400);
+				}
+				const file = imageResponse.data;
+
+				// 2. hash image
+				const hash = crypto.createHash('sha256');
```
notlee, line 86: since this isn't for security maybe using a quicker hash like `md5` would be good here? the benefit might be very minor relatively, so feel free to ignore this 😅

lib/middleware/process-image-request.js
```diff
@@ -47,17 +60,75 @@ function processImage(config) {
...
+				hash.update(file);
 
-		// Store the transform and applied transform for
-		// use in later middleware
-		request.transform = transform;
-		request.appliedTransform = appliedTransform;
-		next();
+				const imageName = hash.digest('hex');
+				transform.setName(imageName);
+				cache[originalImageURI] = imageName;
```
notlee, line 91: how big could the two `cache` objects get? do you think we should be worried about our Heroku provisions?

lib/middleware/process-image-request.js
```diff
@@ -47,17 +60,75 @@ function processImage(config) {
...
+				
+				// 3. check cloudinary to see if image with name exists
+				let imageAlreadyUploadedToCloudinary = false;
+				try {
+					imageAlreadyUploadedToCloudinary = await resource(imageName);
+					// console.log({imageAlreadyUploadedToCloudinary});
```
notlee, line 97: Some commented code to clean up here I think.

lib/middleware/process-image-request.js
```diff
@@ -47,17 +60,75 @@ function processImage(config) {
...
+				} catch (imageAlreadyUploadedToCloudinaryError) {
+					// console.dir({imageAlreadyUploadedToCloudinaryError});
```
notlee, line 99: What happens if this fails silently and we try to upload an image with the same name?

notlee (commented):
Nice work!

This looks good but makes me a little nervous 😅  
before release we should let operations know and be ready for a rollback

Some small feedback/questions in the meantime

lib/middleware/process-image-request.js
JakeChampion: A thousand runs of md5/sha1/sh256:
md5: 3.01ms
sha1: 2.736ms
sha256: 2.397ms

```js
// test.js
const crypto = require("crypto");

function hashTime(algorithm) {
	console.time(algorithm);
	let i = 1000;
	while (i--) {
		crypto.createHash(algorithm).update(String(Math.random()));
	}
	console.timeEnd(algorithm);
}

hashTime("md5");
hashTime("sha1");
hashTime("sha256");
```

lib/middleware/process-image-request.js
JakeChampion: I don't think they will get big as the dynos restart every 24 hours.
We can replace these objects with a Least Recently Used cache which has a max number of items it will cache instead if you want?

lib/middleware/process-image-request.js
JakeChampion: Uploading an image with the same name is fine because we have set the upload config `overwrite` to `true`, which tells cloudinary to overwrite the existing image.

lib/middleware/process-image-request.js
JakeChampion, line 97:
```suggestion
```

lib/middleware/process-image-request.js
JakeChampion, line 99:
```suggestion
```

lib/middleware/process-image-request.js
```diff
@@ -47,17 +60,73 @@ function processImage(config) {
...
+				const imageName = hash.digest('hex');
+				transform.setName(imageName);
+				cache[originalImageURI] = imageName;
+				
+				// 3. check cloudinary to see if image with name exists
+				let imageAlreadyUploadedToCloudinary = false;
+				try {
+					imageAlreadyUploadedToCloudinary = await resource(imageName);
+				} catch (imageAlreadyUploadedToCloudinaryError) {
+				}
```
JakeChampion, line 98:
```suggestion
				} catch (imageAlreadyUploadedToCloudinaryError) {
					// errors will be thrown if either the resource does not exist or if the cloudinary api rate-limit has been reached.
					// both of those errors are safe to ignore because the upload call below does not have a rate-limit.
				}
```

lib/middleware/process-image-request.js
notlee: yeah see I really think we should be optimising for that ~0.060s per million ;)

notlee (approved):
approved! (pending failing test)

PR #627, opened 2020-09-15
stop adding client errors to Sentry as these are not solvable by the application

JakeChampion:
This will make our Sentry alerts less noisy as we should only be receiving Sentry alerts for issues within the application rather than alerts for when a client provided image url is a 404/500/not-an-image.


chee (approved):
these will still be captured in the image service logs though, right?

JakeChampion:
> these will still be captured in the image service logs though, right?

Yes, the logs in Splunk will still have them recorded. Do you think that's the right thing to do?

PR #628, opened 2020-09-15
Make the cloudinary upload process faster by avoiding the resource api call

JakeChampion:
We made a call to the Cloudinary Resource API to check if the image was already uploaded. It turns out that API call is not required as the Upload API call does not upload the image if it already exists in Cloudinary.

Using the Upload API with `overwrite` set to `false` will upload the image to cloudinary if it does not already exist. If it does already exist it returns an object which has the metadata for the already uploaded image.

This has a side-benefit of not using an API call which is rate-limited to 5000 calls per hour, which we regularly were hitting


chee (approved):
github says `const resource` is no longer needed. i can't provide a suggestion beause the definition is not in the diff

PR #629, opened 2020-09-15
Increase caching time of client misconfiguration errors

JakeChampion:
These errors are all triggered by creating an invalid Image Service URL which means that they would always be a client error and are safe to cache indefinitely.

To fix these errors the client would need to update their URL, which means it is a different object in our CDN cache.


chee (approved):
very nice :+1:

PR #630, opened 2020-09-15
increase the timeout for downloading the original image to 20 seconds

JakeChampion:
we have 30 seconds to deal with a request in heroku, let's use at most 20 of those seconds downloading the original image. Hopefully this will reduce the amount of time-out errors we are throwing.

screenshot of sentry which shows 2700 time-out errors in the last 6 days


chee (approved):
:)

PR #631, opened 2020-09-15
reduce load on origin servers by increasing the cache for bad responses

JakeChampion:
When the original image returns a 4xx or higher, let's cache the response for 5 minutes to avoid being overloaded by loads of requests for that image.


chee (approved):
looks good

PR #632, opened 2020-09-16
Cache bad responses due to expired/invalid certificates for 5 minutes

JakeChampion:


notlee (approved):
Do we have any alerts in place for an unexpected number of timeouts of similar errors?

JakeChampion:
> Do we have any alerts in place for an unexpected number of timeouts of similar errors?

Not yet for timeouts, improving our alerts is what I am currently looking at 👍

PR #633, opened 2020-09-16
Improve handling of images which are too large to upload to Cloudinary

JakeChampion:
if the image is too large to upload, let's cache the error response for 5 minutes to help avoid load on the Heroku applications and let's not capture the error in Sentry because it is a client error.

### DIFF
--- a/lib/middleware/process-image-request.js
+++ b/lib/middleware/process-image-request.js
@@ -96,13 +96,22 @@ function processImage(config) {
 				cache[originalImageURI] = imageName;
 				
 				// 3. upload the image to cloudinary - this will not error if the image has already been uploaded to cloudinary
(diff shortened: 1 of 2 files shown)


lib/middleware/process-image-request.js
```diff
@@ -152,7 +161,7 @@ function processImage(config) {
 			request.appliedTransform = appliedTransform;
 			next();
 		}).catch(error => {
-			console.error(error);
+			console.error({error});
```
JakeChampion, line 164:
```suggestion
			console.error(error);
```

notlee (approved):
shame there's no error code?

JakeChampion:
> shame there's no error code?

Yep, it's rather annoying that they don't send one

PR #637, opened 2020-09-18
Fix image requests where only one slash was supplied after `http:`

JakeChampion:
We had previously only fixed this type of mistake for http url which resolved to an asset stored in UPP and not for all http urls.


notlee:
Should we fix this kind of mistake? An error seems reasonable

JakeChampion:
We already fix it for some http URIs which is why I think we should fix it for them all

notlee:
> We already fix it for some http URIs which is why I think we should fix it for them all

suppose that makes sence 👍

notlee (approved):

PR #639, opened 2020-09-18
Cache more errors in fastly and do not record them in sentry

JakeChampion:
Handling errors for:
- dns timeouts - cached for 30 seconds because this is recoverable based on traffic load to origin.

- unreachable networks - cached for 30 seconds because the network should come back online.

- invalid certificates which causes the request to fail - cached for a shortish duration because the certificate could get updated and the request would then work again.

- invalid characters in a url -- these can be cached for a very long time because it will always return an error.

### DIFF
--- a/test/unit/lib/middleware/process-image-request.test.js
+++ b/test/unit/lib/middleware/process-image-request.test.js
@@ -231,6 +231,9 @@ describe('lib/middleware/process-image-request', () => {
 						syscall: 'syscall',
 						code: 'ECONNRESET',
 					});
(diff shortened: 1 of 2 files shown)


test/unit/lib/middleware/process-image-request.test.js
```diff
@@ -6,7 +6,7 @@ const sinon = require('sinon');
 const httpMock = require('node-mocks-http');
 const nock = require('nock');
 
-describe('lib/middleware/process-image-request', () => {
+describe.only('lib/middleware/process-image-request', () => {
```
notlee, line 9:
```suggestion
describe('lib/middleware/process-image-request', () => {
```

notlee (approved):

PR #640, opened 2020-09-18
Add an development mode which has huge stacktraces across async contexts to help debug errors.

JakeChampion:
I've been using this for a while to help debug errors which were in sentry


notlee (approved):

notlee:
approved but would be great to add more details to the readme

PR #641, opened 2020-09-21
Redirect old blogs domain images to the new blogs domain

JakeChampion:
The old blogs domain is still being used for images and the domain no longer exists. We should replace the old domain with the new domain to make these images still work.

Currently we get roughly 4 thousand image requests for the old domain each day.

### DIFF
--- a/lib/routes/v2/images.js
+++ b/lib/routes/v2/images.js
@@ -27,6 +27,9 @@ module.exports = app => {
 		next();
 	}
 
+	// Redirect the old now deleted domain to the correct domain for Live Blogs Images.
+	// Conversation in Slack about this -- https://financialtimes.slack.com/archives/CPEETMY49/p1600188171025000
+	// System in Biz-Ops -- https://biz-ops.in.ft.com/System/blogs
 	function updateOldBlogsLinks(request, response, next) {
 		if (request.params.schemeUrl.startsWith('https://blogs.r.ftdata.co.uk/')) {
 			request.params.schemeUrl = request.params.schemeUrl.replace('https://blogs.r.ftdata.co.uk', 'https://blogs.ft.com');
--- a/lib/routes/v2/images.js
+++ b/lib/routes/v2/images.js
@@ -27,6 +27,17 @@ module.exports = app => {
 		next();
 	}
 
+	function updateOldBlogsLinks(request, response, next) {
+		if (request.params.schemeUrl.startsWith('https://blogs.r.ftdata.co.uk')) {
+			request.params.schemeUrl = request.params.schemeUrl.replace('https://blogs.r.ftdata.co.uk', 'https://blogs.ft.com');
+			request.params.imageUrl = request.params.schemeUrl;
+		} else if (request.params.schemeUrl.startsWith('http://blogs.r.ftdata.co.uk')) {
+			request.params.schemeUrl = request.params.schemeUrl.replace('http://blogs.r.ftdata.co.uk', 'https://blogs.ft.com');
+			request.params.imageUrl = request.params.schemeUrl;
+		}
+		next();
+	}
+
 	function mapScheme(request, response, next) {
 		request.params.scheme = request.params[2] || 'http'; // We make protocol-relative links use http.
 		next();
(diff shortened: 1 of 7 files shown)


lib/routes/v2/images.js
```diff
@@ -27,6 +27,17 @@ module.exports = app => {
 		next();
 	}
 
+	function updateOldBlogsLinks(request, response, next) {
+		if (request.params.schemeUrl.startsWith('https://blogs.r.ftdata.co.uk')) {
```
JakeChampion, line 31:
```suggestion
		if (request.params.schemeUrl.startsWith('https://blogs.r.ftdata.co.uk/')) {
```

lib/routes/v2/images.js
```diff
@@ -27,6 +27,17 @@ module.exports = app => {
...
+			request.params.schemeUrl = request.params.schemeUrl.replace('https://blogs.r.ftdata.co.uk', 'https://blogs.ft.com');
+			request.params.imageUrl = request.params.schemeUrl;
+		} else if (request.params.schemeUrl.startsWith('http://blogs.r.ftdata.co.uk')) {
```
JakeChampion, line 34:
```suggestion
		} else if (request.params.schemeUrl.startsWith('http://blogs.r.ftdata.co.uk/')) {
```

notlee:
Do you know where the requests are coming from? I'm wondering if there's one main source of outdated requests we could update? I'm not convinced mapping old domains should be the image services responsibility.

JakeChampion:
> Do you know where the requests are coming from? I'm wondering if there's one main source of outdated requests we could update? I'm not convinced mapping old domains should be the image services responsibility.

They come from a bunch of old blog articles which are not going to be worked on. I already checked to see if we could get them updated at the source and it looks like no because they are decommissioning that system soon.

test/integration/v2/images-raw.test.js
```diff
@@ -31,11 +31,25 @@ const testImageUris = {
...
+	oldLiveBlogsDomainHttp: 'http://blogs.ft.com/tech-blog/files/2012/03/Screen-Shot-2012-03-01-at-11.25.02-PM-391x270.png',
+	oldLiveBlogsDomainHttps: 'https://blogs.ft.com/tech-blog/files/2012/03/Screen-Shot-2012-03-01-at-11.25.02-PM-391x270.png',
 };
 
 describe('GET /v2/images/raw…', function() {
 
+	describe('/http://blogs.r.ftdata.co.uk', function() {
+		setupRequest('GET', `/v2/images/raw/${testImageUris.oldLiveBlogsDomainHttp}?source=test`);
+		itRespondsWithStatus(200);
+		itRespondsWithContentType('image/png');
```
JakeChampion, line 44:
```suggestion
		itRespondsWithContentType('image/jpeg');
```

test/integration/v2/images-raw.test.js
```diff
@@ -31,11 +31,25 @@ const testImageUris = {
...
+	});
+
+	describe('/https://blogs.r.ftdata.co.uk', function() {
+		setupRequest('GET', `/v2/images/raw/${testImageUris.oldLiveBlogsDomainHttps}?source=test`);
+		itRespondsWithStatus(200);
+		itRespondsWithContentType('image/png');
```
JakeChampion, line 50:
```suggestion
		itRespondsWithContentType('image/jpeg');
```

notlee:
>They come from a bunch of old blog articles which are not going to be worked on. I already checked to see if we could get them updated at the source and it looks like no because they are decommissioning that system soon.

Thanks for the context. I'm guessing these have started 404ing since we began removing images from Cloudinary's storage? Could we document what system this is for and maybe create an issue to remove the redirect when the system is decommissioned? Although I guess there's the chance new systems will come to depend on the redirect.

JakeChampion:
> Thanks for the context. I'm guessing these have started 404ing since we began removing images from Cloudinary's storage? Could we document what system this is for and maybe create an issue to remove the redirect when the system is decommissioned? Although I guess there's the chance new systems will come to depend on the redirect.

What do you think of this? https://github.com/Financial-Times/origami-image-service/pull/641/files#diff-12383274d6f316ab2f3f636db95a140bR30-R32

JakeChampion:
> Thanks for the context. I'm guessing these have started 404ing since we began removing images from Cloudinary's storage? Could we document what system this is for and maybe create an issue to remove the redirect when the system is decommissioned? Although I guess there's the chance new systems will come to depend on the redirect.

Yep, if this is accepted and merged we will create an issue to figure out how to decommission it. I have a splunk search which reports how many of these requests come through, when that number drops dramatically we can have it alert us to decommission this hack.

notlee (approved):
sounds great 👍

PR #643, opened 2020-09-21
 Add error handling for misformatted hostnames and no name-servers for a domain

JakeChampion:
- Misformatted hostnames can be cached for a long time because it is a client issue in the hostname and the hostname is part of the cache-key we use in Fastly, the only waay the client can fix the issue is to change their hostname, which would mean a new cache-key is used.

- Domains which have no name-server configured can be fixed by the client without changing anything which modifies our cache-key, so we need to use a short cache time.

- I've removed the check for ENOTFOUND errors to have been thrown by a system call because we no longer use the system dns implementation. Our new implementation does not add a `syscall : 'getaddrinfo'` property to the errors.

### DIFF
--- a/lib/image-service.js
+++ b/lib/image-service.js
@@ -48,7 +48,7 @@ function createProxy(errorHandler) {
 	proxy.on('proxyReq', proxyRequestHandler);
 	proxy.on('proxyRes', proxyResponseHandler);
 	proxy.on('error', (error, request, response) => {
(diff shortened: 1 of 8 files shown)


notlee (approved):
nice work 👍 

are we duplicating logic for error handling in the svg middleware? If so maybe we could share that logic?

JakeChampion:
> nice work 👍
> 
> are we duplicating logic for error handling in the svg middleware? If so maybe we could share that logic?

That's going to happen soon 👍

PR #646, opened 2020-09-24
Extract the axios error handling into it's own module for reuse

JakeChampion:
These errors are thrown and handled in at least three different parts of the codebase:
- The part which handles tinting SVGs.
- The part which converts ftcms images to their http url
- The part which download the image and then uploads it to cloudinary.

They all handle these errors in the same way, instead of copying and pasting the error handing code between them, let's put it in a module to reuse it.

### DIFF
--- a/lib/create-error-from-axios-error.js
+++ b/lib/create-error-from-axios-error.js
@@ -1,0 +1,76 @@
+'use strict';
+
+module.exports = function createErrorFromAxiosError(error, address) {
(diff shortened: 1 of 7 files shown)


notlee (approved):
Much improved :D I thought `handleAxiosErrors` is maybe a misleading name though since handle makes me think its doing something to make the error ok, maybe something like `createErrorForAxiosError`...or something?

Issue #657, opened 2020-10-11
Year 2018 is used


krumft:
https://github.com/Financial-Times/origami-image-service/blob/9ea639a8cc2816f98fc3298aa6b5a22e96902bac/views/partials/footer.html#L18

PR #659, opened 2020-10-13
Update year to 2020 in footer

JakeChampion:
Fixes https://github.com/Financial-Times/origami-image-service/issues/657

### DIFF
--- a/views/partials/footer.html
+++ b/views/partials/footer.html
@@ -15,7 +15,7 @@
 					<a href="http://help.ft.com/help/legal-privacy/privacy/" class="o-footer-services__bulletted-link">Privacy</a>
 					<a href="http://help.ft.com/help/legal-privacy/terms-conditions">Terms & Conditions</a>
 				</div>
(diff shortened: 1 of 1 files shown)


chee (approved):
isn't the copyright meant to be the year of first publication?

are we really claiming that these works never enter the public domain because we occasionally file pull requests?

if only Walt had thought of semver'ing Mickey Mouse, moving a whisker 1pt every year and calling it a new Mickey Mouse


BREAKING CHANGES IN THIS MICKEY MOUSE:
- steamboat api removed (has been deprecated since v2, now fully removed)

JakeChampion:
> isn't the copyright meant to be the year of first publication?

I have no idea, www.ft.com also list the copyright year as 2020 🤷

krumft:
> > isn't the copyright meant to be the year of first publication?
> 
> I have no idea, [www.ft.com](http://www.ft.com) also list the copyright year as 2020 🤷

I also don't have idea what this year should be from legal standpoint.

PR #689, opened 2020-12-22
Add a /placeholder/ api for generating placeholder images

### LINKED ISSUE #339: Proposal: Add a placeholder images API
# Background
Have you ever worked with src-set? I guess so. If you did, you know that sometimes can be tricky especially when you have to achieve complex layouts and maybe you have to use sizes...oh gosh.
That is frustrating, time-consuming and error-prone, especially for junior devs.
The following proposal is meant to facilitate visual feedbacks in dev mode while devs are working on src-set and can potentially be used also for visual tests that involve screen snapshots.

# How does it work?
Given the background, let's say I'm working in my app implementing src-set.
In my 'component-image', that is my centralised point where I generate the URLs for images using the origami-image-service, I enable an option that adds a query parameter, something like...
```
...&debug=1...
```
So now, my src url that previously was something like...
```
&width=150&quality=medium&dpr=1
```
is ...
```
...&width=150&quality=medium&dpr=1&debug=1...
```
Origami-image-service receiving these URLs with the debug option enabled, regardless whatever images I'm requesting, should be capable to produce a mocked-up image similar to what services like [placeholder.com](https://placeholder.com/) do where I can see the parameters of the image I have requested directly in the browser...

screen shot 2018-11-01 at 10 31 21

The image will report all the information related to the image so I can verify with zero effort that my implementation is working correctly.
In this way, I won't need to open the dev tools network panel to check what image I'm requesting.

# Does Cloudinary provide this capability already?
I had a quick look but found anything, if we want to proceed we should probably better investigate if they offer something similar, otherwise, I'm pretty sure will be simple implement the feature via an image library without reaching Cloudinary at all for these kinds of images.

# Everything just for src-set?
Well no, potentially the feature could be used also for Visual tests where you need to collect snapshots to use as references against the tests.
That's would mean a simple and more reliable way to cover src-set functionality.

# Why in the image service?
Centralised it here make avoid the creating of this functionality in a distributable package that every application should use it and that should mock-up also the image-service API that can change.
I think have it in the service is the simplest and more efficient way to implement this.

chee:
closes #339

here's an example

an image showing width, height and quality values as text

it generates them in different o-colors, isn't it nice


chee:
nice, the tests are passing.
coverage is lower, though, i haven't written any tests for the new code

chee:
i suppose we might want to write documentation for this

lib/get-color-pair.js
```diff
@@ -0,0 +1,40 @@
+'use strict';
+
+const colors = {
```
JakeChampion, line 3: I've added a comment which mentions that the colors come from o-colors --
```suggestion
// Currently hard-coding the o-colors values
// We should use o-colors values directly when that is possible.
// Currently waiting on https://github.com/Financial-Times/o-colors/issues/289
const colors = {
```

JakeChampion (commented):
Looking good, we need to update the documentation on the website to mention this new feature.

Should we also add it to the url builder page?

chee:
> Looking good, we need to update the documentation on the website to mention this new feature.
> 
> Should we also add it to the url builder page?

we should write about in the API docs for sure.
how were you pictureing it in the url builder?

JakeChampion:
> how were you pictureing it in the url builder?

We could add a radio group which has a radio input for each mode we want to list. We could list a subset of the modes such as: image, and placeholder. The other modes don't return images so may not be useful to have in the radio group

JakeChampion:
> > how were you pictureing it in the url builder?
> 
> We could add a radio group which has a radio input for each mode we want to list. We could list a subset of the modes such as: image, and placeholder. The other modes don't return images so may not be useful to have in the radio group

Let's keep the url-builder as a separate feature which we could add in the future. The only thing this pull-request is waiting on to be merge-able is updating the documentation.

JakeChampion (approved):

PR #748, opened 2021-04-20
Update to node 14 LTS and npm 7

JakeChampion:
Origami spec 2.0 components require npm 7.

I figured it might be good to get our services onto npm 7 in preparation for the components updates


JakeChampion:
If we accept this, we should watch the application metrics after we deploy to production, in the past upgrading the node version has unearthed memory leaks

notlee (approved):

PR #752, opened 2021-04-29
Switch to build-service v3!

chee:
it works!

notes:
- the error message around duplicate entries in the `components=` array should mention what is duplicated
- having to add `@financial-times/` to every component name is a drag


notlee:
Looks good! Let's not release this until obs v3 is finalised? Nice this work brought up those two notes :)

JakeChampion (approved):

PR #762, opened 2021-09-28
Enable the integration tests to run against a live server

JakeChampion:
We need this functionality to have tests written for the new way of serving image-sets (#763)

As the image-sets will be served from the application instead of a separate domain, the integration tests for image-sets will only work when the integration test server is on a domain which cloudinary can access (not localhost)

This setup is the same as the setup we have in the origami build service (https://github.com/Financial-Times/origami-build-service/pull/411)

### DIFF
--- a/.github/workflows/deploy-to-dev.yml
+++ b/.github/workflows/deploy-to-dev.yml
@@ -9,3 +9,28 @@ jobs:
       - run: git fetch --prune --unshallow
       - name: Deploy to Heroku
         run: git push https://heroku:${{ secrets.HEROKU_AUTH_TOKEN }}@git.heroku.com/origami-image-service-dev.git HEAD:refs/heads/master --force
+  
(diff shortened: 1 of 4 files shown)


.github/workflows/deploy-to-dev.yml
```diff
@@ -9,3 +9,28 @@ jobs:
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: Financial-Times/change-api-action@v1
+        name: Create new change log
+        with:
+          change-api-key: ${{ secrets.CHANGE_API_KEY }}
+          system-code: "origami-image-service-v2"
+          environment: dev
+          slack-channels: "ft-changes,origami-deploys"
+
```
chee, line 25: :o

chee (approved):
this seems great to me

PR #763, opened 2021-09-28
move the imagesets into the repository instead of installing via npm

JakeChampion:
Most image-sets are used only via the origami-image-service. Moving the image-sets from their own repositories into this repository will greatly reduce the amount of steps require to update an image in an imageset and see the change appear on the origami-image-service.

Still remaining:
- [ ] Add documentation about imagesets on the image-service website
- [ ] Add json endpoint to return images in an imageset (this will be used to update fticons/o-icons in the origami repository)


JakeChampion:
> I may have ruined this PR by making a `logo-images` release with a new image

I need to make some changes to the PR anyway. Thank you for mentioning the new version of logo-images, I could have missed that!

PR #767, opened 2021-11-04
fix: url builder output includes double forward slash

notlee:
we can do a bit better


JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v451.2.0 :tada:

PR #768, opened 2021-11-04
fix: url builder output in production where the hostname env is not set

notlee:
Express gets `request.hostname` from the Host HTTP header, I think it's safe to use in this way


notlee:
This small issue is not going well for me today

JakeChampion (approved):

origamiserviceuser:
:tada: This PR is included in version v451.2.1 :tada:

PR #782, opened 2022-02-16
fix: cache ftcms requests which 404 only for 5m instead of 1 year

JakeChampion:
Recently the content api had an issue where it was returning 404s by
accident for requests which should have returned a 200. We then cached
those invalid responses for 1 year, which was unfortunate for our users
as when the content api was fixed, we were still serving 404s.

This change will make it so that we cache content api 404s for only 5
minutes, which is long enough to stop our servers getting overloaded
and is short enough for our users to retry the requests in case the 404
was an accident


rowanbeentje (approved):
When I was wondering about cache times I was imagining something around the half hour mark but I imagine these CMS checks are fairly cheap anyway! 👍

origamiserviceuser:
:tada: This PR is included in version v451.2.4 :tada:

PR #783, opened 2022-02-16
Always add the origami-image-service surrogate key to images api responses

JakeChampion:


rowanbeentje (approved):
So this runs well before the existing surrogate key logic (https://github.com/Financial-Times/origami-image-service/blob/master/lib/image-service.js#L171-L178), acting as a base value, and then actual image responses will overwrite that base value with the normalised + encoded keys?

If my understanding is correct, looks good!

origamiserviceuser:
:tada: This PR is included in version v451.2.5 :tada:

JakeChampion:
> So this runs well before the existing surrogate key logic (https://github.com/Financial-Times/origami-image-service/blob/master/lib/image-service.js#L171-L178), acting as a base value, and then actual image responses will overwrite that base value with the normalised + encoded keys?
> 
> If my understanding is correct, looks good!

You are correct ☺️

PR #787, opened 2022-03-28
attach all route handlers under the path prefix /__origami/service/image/

JakeChampion:
this is because the production origami image service is served under www.ft.com/__origami/service/image/ and having the application server respond to this path will simplify our cdn configuration

By making this change we can remove these parts of the cdn configuration:
https://github.com/Financial-Times/ft.com-cdn/blob/92a50479428e46b35a63485d100a8694e448c327/src/vcl/origami-image-service.vcl#L16
and 
https://github.com/Financial-Times/ft.com-cdn/blob/92a50479428e46b35a63485d100a8694e448c327/src/vcl/origami-image-service.vcl#L67-L76

The reason we have added new routes instead of replacing the existing routes is because when this is deployed it will need to respond to the old routes whilst we make the changes to the CDN to remove the `origami_image_service_miss` and `origami_image_service_pass` subroutines which remove the path prefix /__origami/service/image/ before making the request to our application servers

### DIFF
--- a/data/navigation.json
+++ b/data/navigation.json
@@ -2,32 +2,32 @@
 	"items": [
 		{
 			"name": "Overview",
-			"href": "v2",
+			"href": "/__origami/service/image/v2",
 			"current": false
 		},
 		{
 			"name": "URL Builder",
-			"href": "v2/docs/url-builder",
+			"href": "/__origami/service/image/v2/docs/url-builder",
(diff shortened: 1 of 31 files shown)


origamiserviceuser:
:tada: This PR is included in version v451.3.0 :tada:

PR #788, opened 2022-03-30
Remove ft-image-format header as it is not used by the application

JakeChampion:


lib/image-service.js
```diff
@@ -182,7 +174,7 @@ function createProxy(errorHandler) {
 			proxyResponse.headers['Content-Dpr'] = request.transform.getDpr();
 		}
 
-		proxyResponse.headers['Vary'] = 'FT-image-format, Content-Dpr, Accept';
```
notlee, line 185: You [mention here](https://github.com/Financial-Times/origami-image-service/pull/790) that we can delete the whole of `origami_image_service_fetch` when this PR is released. Could you expand on why? I thought we'd need to Vary on format and can't see how that would happen

JakeChampion: ft-image-format looks to not be used at all when decided what response to serve which is why it's been removed from the Vary header. We still have Accept in the Vary header and Accept is used to figure out whether to serve webp or jpegxr or the original format of the image.

because the backend application is serving a correct Vary response header, we don't need to have a vcl_fetch implementation which would be adding the same value to the Vary header that is already there

notlee (approved):

PR #789, opened 2022-03-30
Add support for AVIF

JakeChampion:
if the client has `image/avif` in their Accepts header and did not request a specific image format such as svg or png, then we will now serve them an Avif image

webp: 115.19kb
https://origami-image-service-qa.herokuapp.com/__origami/service/image/v2/images/raw/http%3A%2F%2Fim.ft-static.com%2Fcontent%2Fimages%2Fa60ae24b-b87f-439c-bf1b-6e54946b4cf2.img?source=docs
image

avif: 73.31kb
https://origami-image-service-dev.herokuapp.com/__origami/service/image/v2/images/raw/http%3A%2F%2Fim.ft-static.com%2Fcontent%2Fimages%2Fa60ae24b-b87f-439c-bf1b-6e54946b4cf2.img?source=docs
image


lib/image-service.js
```diff
@@ -1,5 +1,12 @@
 'use strict';
 
+const http = require('http');
```
JakeChampion, line 3: We had to upgrade to a newer version of axios because the old version could no longer install on osx using npm.

This is the way to enable dns caching for the newer version of axios

lib/middleware/get-cms-url.js
```diff
@@ -2,7 +2,6 @@
 
 const httpError = require('http-errors');
 const axios = require('axios').default;
-const dnsLookup = require('lookup-dns-cache').lookup;
```
JakeChampion, line 5: The dns caching is now added in a single location in lib/image-service.js

lib/middleware/get-image-meta.js
```diff
@@ -9,16 +9,13 @@ module.exports = getImageMeta;
...
 	if (acceptHeader && acceptHeader.includes('image/jxr')) {
 		acceptHeader = '*/*';
 	}
 
 	// Get the image meta data
-	probe({
-
-		// Request the applied transform
-		url: request.appliedTransform,
+	return probe(request.appliedTransform, {
```
JakeChampion, line 18: the version of probe-image-size we were using did not support avif images so I updated to a version which does. The new version changed their interface from callbacks to promises

lib/middleware/process-image-request.js
```diff
@@ -77,6 +76,10 @@ function processImage(config) {
 			}
 		}
 
+		if (request.get('accept') && request.get('accept').includes('image/avif') && transform.format === 'auto') {
```
JakeChampion, line 79: we only serve an avif image if the accept header explicitly contains `image/avif` and the format to serve the image is set to `auto`

package.json
```diff
@@ -22,18 +22,17 @@
   "main": "./lib/image-service.js",
   "scripts": {
     "start": "heroku-node-settings index.js",
-    "dev": "node --stack_trace_limit=100 -r trace -r clarify index.js",
-    "snyk-protect": "snyk protect",
```
JakeChampion, line 26: I don't think we need snyk installed anymore as the project uses the github snyk integration and doesn't make use of snyk's patches

lib/middleware/get-headshot-url.js
```diff
@@ -2,7 +2,6 @@
 
 const httpError = require('http-errors');
 const axios = require('axios').default;
-const dnsLookup = require('lookup-dns-cache').lookup;
```
JakeChampion, line 5: The dns caching is now added in a single location in lib/image-service.js

lib/middleware/handle-svg.js
```diff
@@ -3,7 +3,6 @@
 const createDOMPurify = require('dompurify');
 const httpError = require('http-errors');
 const axios = require('axios').default;
-const dnsLookup = require('lookup-dns-cache').lookup;
```
JakeChampion, line 6: The dns caching is now added in a single location in lib/image-service.js

lib/middleware/process-image-request.js
```diff
@@ -5,7 +5,6 @@ const ImageTransform = require('../image-transform');
 const cloudinaryTransform = require('../transformers/cloudinary');
 const url = require('url');
 const axios = require('axios').default;
-const dnsLookup = require('lookup-dns-cache').lookup;
```
JakeChampion, line 8: The dns caching is now added in a single location in lib/image-service.js

lib/middleware/process-image-request.js
notlee: Am I right in thinking we're setting a format explicitly for now because cloudinary doesn't serve avif automatically yet? If so would be great to put a comment to that affect so we can remove it confidently in the future

https://cloudinary.com/blog/how_to_adopt_avif_for_images_with_cloudinary#how_do_i_do_that_

notlee (approved):
Nice! We should shout about this before release. I'd like to see the impact on SpeedCurve :D

lib/middleware/process-image-request.js
JakeChampion: Yes, I've added a comment which explains that 👍

PR #790, opened 2022-03-31
Set timing-allow-origin and ft-supress-friendly-error response headers within the application

JakeChampion:
as these are response headers, it makes sense to set them in the application.
setting these in the application make them simpler to test and change as it does not invlove touching the ft.com-cdn

if this is merged and release, we can then remove these lines from the ft.com-cdn -> https://github.com/Financial-Times/ft.com-cdn/blob/98bd0a5722f7cc4280e08a3f7dc1d3df240f966b/src/vcl/origami-image-service.vcl#L57-L58

if [this pr](https://github.com/Financial-Times/origami-image-service/pull/788) is also merged and released, then we can remove the entire `origami_image_service_fetch` subroutine from ft.com-cdn


notlee (approved):
nice

origamiserviceuser:
:tada: This PR is included in version v451.4.0 :tada:

PR #805, opened 2022-08-11
fix: do not apply avif to gif after initial request

notlee:
If we have an image in in memory cache:
https://github.com/Financial-Times/origami-image-service/blob/5254832481ef794909eca717d5f27e75147eebed/lib/middleware/process-image-request.js#L95
Then we don't check if the image is a gif to remove the avif transform:
https://github.com/Financial-Times/origami-image-service/blob/5254832481ef794909eca717d5f27e75147eebed/lib/middleware/process-image-request.js#L175

This is a bit not nice, and could use a refactor, but should
stop the production issue without removing avif support.

### DIFF
--- a/lib/middleware/process-image-request.js
+++ b/lib/middleware/process-image-request.js
@@ -16,7 +16,8 @@ const FileType = require('file-type');
 module.exports = processImage;
 
 function processImage(config) {
-	const cache = Object.create(null);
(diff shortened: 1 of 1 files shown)


akomiqaia (approved):

origamiserviceuser:
:tada: This PR is included in version v453.3.1 :tada:

PR #806, opened 2022-09-23
fix: remove better-lookup

notlee:
We have seen many `ENETUNREACH` errors since Apr 6, when this was released. I'm not sure why. Let's see if removing better-lookup will help.


akomiqaia (approved):

origamiserviceuser:
:tada: This PR is included in version v453.3.2 :tada:

PR #834, opened 2023-06-29
fix: increase stroke thickness on label and label-outline SVGs

frshwtr:


chee (approved):
lookin good

rowanmanning (approved):
I agree 👍

PR #835, opened 2023-06-29
feat: transform images with Cloudinary's 'pad' option

frshwtr:
Enables requests to the API with `fit` set to `pad`. This will allow images to be scaled whilst maintaining the original aspect ratio. Any white space that occurs after the transform will bet set with a colour from `bgcolor`.

Partially completes OR-240. FTEdit would like the background to also be transparent, but this doesn't appear to work without further changes. Releasing this as an increment will allow them to test if this feature meets their requirements whilst we work on the transparency issue.

### DIFF
--- a/lib/image-transform.js
+++ b/lib/image-transform.js
@@ -376,7 +376,7 @@ module.exports = class ImageTransform {
 		if (value === 'transparent') {
 			return ImageTransform.colors.white;
 		}
-		if (!/^#?[0-9a-f]{3,8}$/i.test(value)) {
+		if (!/^#?[0-9a-f]{3,6}$/i.test(value)) {
 			value = colornames(value);
 		}
 		if (!value) {
--- a/lib/image-transform.js
+++ b/lib/image-transform.js
@@ -639,7 +639,8 @@ module.exports = class ImageTransform {
 			'contain',
 			'cover',
 			'fill',
-			'scale-down'
+			'scale-down',
+			'pad'
 		];
 	}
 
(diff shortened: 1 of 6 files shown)


views/api.html
```diff
@@ -121,6 +121,8 @@ <h2 id="query-parameters">Query Parameters</h2>
 					<dd>The image width and height are set to the exact dimensions given in the <code>width</code> and <code>height</code> parameters, potentially stretching the image.</dd>
 					<dt>scale-down</dt>
 					<dd>Similarly to <code>contain</code>, the image dimensions are scaled down to be less than or equal to the corresponding dimensions of the frame, but the image is never enlarged.</dd>
+					<dt>pad</dt>
```
rowanmanning, line 124: I think we should also add this option to the URL builder page after [here](https://github.com/Financial-Times/origami-image-service/blob/86014ad72d4039d26e66148329fb5625f8876748/views/url-builder.html#L129). I think it'd also involve a little weird form validation stuff [here](https://github.com/Financial-Times/origami-image-service/blob/86014ad72d4039d26e66148329fb5625f8876748/lib/routes/v2/docs/url-builder.js#L32).

lib/image-transform.js
```diff
@@ -639,7 +639,8 @@ module.exports = class ImageTransform {
 			'contain',
 			'cover',
 			'fill',
-			'scale-down'
+			'scale-down',
+			'pad'
```
rowanmanning, line 643: There's also a JSDoc param [at the top of this file](https://github.com/Financial-Times/origami-image-service/blob/86014ad72d4039d26e66148329fb5625f8876748/lib/image-transform.js#L32) which is out of date after adding `pad`.

rowanmanning (commented):
Caveated with the fact that I don't think I can run this due to Vault permissions etc, everything looks good. There are a few places where the change is missing, though.

frshwtr:
Thanks @rowanmanning. I've resolved both.

akomiqaia (approved):

PR #839, opened 2023-08-24
Update Twitter logo [OR-295]

notlee:
- Replace the "twitter" logo with the new X logo. This will allow us to roll out the brand update semi-automatically. However since this is a rebrand that includes a name change it's not without risk. Though we haven't found any, there may be cases we want to retain the old Twitter logo. We have the ability to roll back this change.
- Deprecate the "twitter" image and introduce a new "x" image, even though both show the same logo, so the image name and branding align.


frshwtr (approved):
With the deprecation of 'twitter', do we also need to update our references to it in [o-share](https://github.com/Financial-Times/origami/blob/main/components/o-share/src/tsx/shareIcon.tsx)?

notlee:
👍 There's an open PR for that @frshwtr https://github.com/Financial-Times/origami/pull/1239

origamiserviceuser:
:tada: This PR is included in version v453.6.0 :tada:

PR #843, opened 2023-09-13
fix: scripts/delete-old-images-from-cloudinary.js [OR-298]

notlee:
To reduce Cloudinary costs we want to delete old images. We have a script to do this, which is run every day via a "Heroku scheduler" on origami-image-service-eu. However this script errors due to api limits, it's stuck trying to delete the same images on loop.

- Pass `next_cursor` to Cloudinary search, so the next results are returned correctly.
- Set `keep_original` when deleting images. This will delete only derived images – transforms – not the originals. We don't promise to keep originals and expect the source images to remain available, but I don't want to have to explain that to any angry product teams right now.
- Update logs to help us monitor the script. If it deletes many images and then fails before the script exits, we want to know.

We probably don't want to run this updated script every day as it will have to process every image we have stored, that is older than a month, everyday. This is because it no longer deletes the originals. Let's run this inititally to clear the backlog of old transforms, monitor the duration of future runs, and consider changing the search expression to process a limited range of images such as those uploaded between the last 3-6 months `.expression('uploaded_at:[24w TO 12w]')`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16,7 +16,7 @@
         "@sentry/node": "^5.30.0",
         "axios": "^0.26.1",
         "base-64": "0.1.0",
-        "cloudinary": "1.22.0",
+        "cloudinary": "1.40.0",
         "colornames": "1.1.1",
         "date-fns": "2.15.0",
         "dompurify": "2.2.2",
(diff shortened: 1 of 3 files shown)


frshwtr (approved):

origamiserviceuser:
:tada: This PR is included in version v453.9.0 :tada:

PR #845, opened 2023-09-26
backstage: validate uuid before making requests to internal apis

akomiqaia:
We are using [uuid](https://www.npmjs.com/package/uuid) to validate img uuids passed to the image service. It seems that all our internal APIs use the same structure for generating UUIDs and this library should be enough to validate if img UUIDs are valid or not. 

From manually testing and making a few requests to image service seems to work and we return 400 before we make malicious requests to internal APIs.

### DIFF
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -22,7 +22,6 @@ function getCmsUrl(config) {
 		const cmsId = uriParts.shift();
 		const query = (uriParts.length ? '?' + uriParts.join('?') : '');
     if (!uuidValidate(cmsId)) {
-      // If UUID is not valid, we error
       const error = httpError(400, `Image key ${cmsId} is invalid`);
       error.cacheMaxAge = '5m';
       error.skipSentry = true;
--- a/lib/middleware/get-cms-url.js
+++ b/lib/middleware/get-cms-url.js
@@ -3,6 +3,7 @@
 const httpError = require('http-errors');
 const axios = require('axios').default;
 const createErrorFromAxiosError = require('../create-error-from-axios-error');
+const {validate: uuidValidate} = require('uuid');
 
 module.exports = getCmsUrl;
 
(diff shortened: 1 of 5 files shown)


lib/middleware/get-cms-url.js
```diff
@@ -20,6 +21,13 @@ function getCmsUrl(config) {
 		const uriParts = request.params.imageUrl.split(':').pop().split('?');
 		const cmsId = uriParts.shift();
 		const query = (uriParts.length ? '?' + uriParts.join('?') : '');
+    if (!uuidValidate(cmsId)) {
+      // If UUID is not valid, we error
```
frshwtr, line 25: I think the code is descriptive enough to not need a comment here.

test/unit/lib/middleware/get-cms-url.test.js
```diff
@@ -39,14 +57,14 @@ describe('lib/middleware/get-cms-url', () => {
 
 		describe('middleware(request, response, next)', () => {
 			let scope;
-			const v2Uri = 'https://prod-upp-image-read.ft.com/mock-id1';
+			const v2Uri = `https://prod-upp-image-read.ft.com/${uuids.mock1}`;
 
 			beforeEach(done => {
-				origamiService.mockRequest.params.imageUrl = 'ftcms:mock-id1';
+				origamiService.mockRequest.params.imageUrl = `ftcms:${uuids.mock1}`;
```
frshwtr, line 63: Instead of using two different generated urls, could a uuid be generated in the `beforeEach` and stored in a variable. This will give the same behaviour of unique IDs and use existing test structure:

```js
let uuid;
beforeEach(() => {
...
}

it('test1', () => {
    const request = `https://something.a/${uuid}`;
    ....
}

it('test2', () => {
    const request = `https://something.a/${uuid}`;
    ....
}
```

test/unit/lib/middleware/get-cms-url.test.js
```diff
@@ -2,6 +2,24 @@
 
 const assert = require('proclaim');
 const nock = require('nock');
+const {v4: generateUuid} = require('uuid');
+
+
+const uuids = {
```
frshwtr, line 8: What do each of these mocks represent? Instead of defining them in an object, defining each individually in their respective describe blocks will help add more context about the uuid scenario.

frshwtr (approved):

PR #848, opened 2023-10-20
feat: overlay images.

biggest-dave:
Run locally and hit with: 
http://localhost:8080/__origami/service/image/v2/images/raw/https%3A%2F%2Fd1e00ek4ebabms.cloudfront.net%2Fproduction%2Fa88c2b59-ac2d-4f70-a68a-57b8499bd70f.jpg?source=next-article&fit=scale-down&quality=highest&width=700&dpr=1&overlay=https://ft-next-assets-prod.s3.eu-west-1.amazonaws.com/assets/ft-banner.png&overlay_height=0.5&overlay_y=30

You should see: 

Screenshot 2023-10-26 at 10 01 37

### DIFF
--- a/lib/image-transform.js
+++ b/lib/image-transform.js
@@ -63,6 +63,7 @@ module.exports = class ImageTransform {
 			this.setOverlayX(properties.overlay_x);
 			this.setOverlayY(properties.overlay_y);
 			this.setOverlayGravity(properties.overlay_gravity);
+			this.setOverlayCrop(properties.overlay_crop);
 		}
 	}
 
@@ -184,7 +185,26 @@ module.exports = class ImageTransform {
 	 */
 
 	setOverlayGravity(value) {
-		this.overlay_gravity = value ? value : 'south_east';
+		const errorMessage = `Overlay gravity must be one of ${ImageTransform.validOverlayGravities.join(', ')}`;
+		this.overlay_gravity = ImageTransform.sanitizeEnumerableValue(value, ImageTransform.validOverlayGravities, errorMessage);
+	}
+
+	/**
+	 * Get the overlay crop.
+	 * @return {String}
+	 */
+	getOverlayCrop() {
+		return this.overlay_crop;
+	}
+
+	/**
+	 * Set the overlay crop
+	 * @param {String} value
+	 */
+
+	setOverlayCrop(value) {
+		const errorMessage = `Overlay crop must be one of ${ImageTransform.validOverlayCrops.join(', ')}`;
+		this.overlay_crop = ImageTransform.sanitizeEnumerableValue(value, ImageTransform.validOverlayCrops, errorMessage);
 	}
 
 
@@ -771,6 +791,37 @@ module.exports = class ImageTransform {
 		return [
 			'faces',
 			'poi'
+		];
+	}
+
+	/**
+	 * @private
+	 */
+	static get validOverlayGravities() {
+		return [
+			'south_east',
+			'north_west',
+			'south_west',
+			'north_east'
+		];
+	}
+
+	static get validOverlayCrops() {
+		return [
+			'scale',
+			'fill',
+			'crop',
+			'fill_pad',
+			'fit',
+			'imagga_crop',
+			'imagga_scale',
+			'lfill',
+			'limit',
+			'lpad',
+			'mfit',
+			'mpad',
+			'pad',
+			'thumb'
 		];
 	}
 
--- a/lib/image-transform.js
+++ b/lib/image-transform.js
@@ -34,13 +34,11 @@ module.exports = class ImageTransform {
 	 * @param {String} [properties.format=auto] - The file format of the transformed image. One of auto, jpg, png, svg.
 	 * @param {String} [properties.bgcolor] - A background color to apply to the transformed image (if transparent). Hex code or named color.
 	 * @param {String} [properties.overlay] - THE URI of an overlay image.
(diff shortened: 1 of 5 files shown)


lib/transformers/cloudinary.js
```diff
@@ -50,11 +50,42 @@ function buildCloudinaryTransforms(transform) {
...
-	return cloudinaryTransforms;
+
+	const overlay = transform.getOverlay();
+
+
+	if (overlay) {
+
+		const overlayPosition = transform.getOverlayPosition();
+		const overlayHeight = transform.getOverlayWidth();
+		const overlayWidth = transform.getOverlayHeight();
```
notlee, line 66: Oops! Even when I found the problem, it took me so long to see that this was the cause 🙈 

```suggestion
		const overlayHeight = transform.getOverlayHeight();
		const overlayWidth = transform.getOverlayWidth();
```

lib/image-transform.js
```diff
@@ -68,6 +87,147 @@ module.exports = class ImageTransform {
...
+	}
+
+	/**
+	 * Set the overlay width
+	 * @param {String} value 
+	 */
+
+	setOverlayWidth(value) {
+		// the cloudinary api will ignore an empty string
+		this.overlay_width = value ? value : '';
```
notlee, line 138: Reviewing your PR is getting me more (re?) familiarised with the Image Service which is good 😄 Looks like `lib/image-transform.js` is intended as an abstraction layer. Should Cloudinary specific fallbacks be moved to `lib/transformers/cloudinary.js`?

lib/image-transform.js
```diff
@@ -68,6 +87,147 @@ module.exports = class ImageTransform {
...
+		 * Get the overlay x.
+		 * @return {String}
+		 */
+	getOverlayX() {
+		return this.overlay_x;
+	}
+
+	/**
+	 * Set the overlay x
+	 * @param {String} value
```
notlee, line 170: Could/should this be `{string|number}`?

lib/image-transform.js
```diff
@@ -49,6 +57,17 @@ module.exports = class ImageTransform {
...
+		
+		if (properties.overlay) {
+			this.setOverlay(properties.overlay);
+			this.setOverlayPosition(properties.overlay_position);
+			this.setOverlayWidth(properties.overlay_width);
+			this.setOverlayHeight(properties.overlay_height);
+			this.setOverlayX(properties.overlay_x);
+			this.setOverlayY(properties.overlay_y);
+			this.setOverlayGravity(properties.overlay_gravity);
+			this.setOverlayCrop(properties.overlay_crop);
```
notlee, line 69: For each of these, could we limit options here to a known set? Maybe even just the options you need right now? See `setGravity`. That way we have a smaller, more well defined task if we opt to move to a new supplier next year (but it's also quite easy to allow extra Cloudinary features).

lib/image-transform.js
```diff
@@ -68,6 +87,147 @@ module.exports = class ImageTransform {
 		this.uri = ImageTransform.sanitizeUriValue(value, errorMessage);
 	}
 
+		/**
+	 * Get the overlay.
+	 * @return {String}
+	 */
+		getOverlay() {
+			return this.overlay;
```
notlee, line 95: Little indenting issue here (urgh, we need to set up a code formatter!)

notlee (changes requested):
Looking good! One little bug and couple of suggestions here with our future maintenance in mind 🙏

notlee:
Also, bonus points! Apologies if I didn't think to mention before. Cache can be tricky with the image service. I suspect it would be a welcome ability to be able to "purge all images with x as their overlay" when we update an overlay's source image. To do that we'll want to set a [Surrogate Key](https://developer.fastly.com/reference/http/http-headers/Surrogate-Key/) on responses for Fastly.

I think that would happen as [part of the proxy response handler](https://github.com/Financial-Times/origami-image-service/blob/b1655858b90bf1455564f734431686030e5ef5e0/lib/image-service.js#L179). We'd want to differentiate an image used as an overlay vs. the same image being requested directly, so we'd maybe want to add something like `encodeKeys('overlay-' + overlayUrl)` conditionally, when an overlay is requested.

lib/transformers/cloudinary.js
biggest-dave: Well spotted!

lib/image-transform.js
biggest-dave: I took out position and crop. I left gravity in as I could image that being used very soon (we're having debates about if it should come in from the south_east or south_west).

notlee: Then we can document the more limited set of values explicitly, rather than need people to reference Cloudinary docs.

biggest-dave: Done!

lib/image-transform.js
```diff
@@ -654,6 +794,37 @@ module.exports = class ImageTransform {
...
+			'fit',
+			'imagga_crop',
+			'imagga_scale',
+			'lfill',
+			'limit',
+			'lpad',
+			'mfit',
+			'mpad',
+			'pad',
+			'thumb'
```
notlee, line 824: Can we reduce this list to just what you need for now? It'll make our potential transition to a new supplier easier

biggest-dave: done

notlee (approved):
Looks good! 🥳 🎉

origamiserviceuser:
:tada: This PR is included in version v453.10.0 :tada:

PR #853, opened 2023-12-13
backstage: add redis integration for monitoring

akomiqaia:


README.md
```diff
@@ -29,7 +29,15 @@ Before we can run the application, we'll need to install dependencies:
 npm install
 ```
 
-Run the application in development mode with
+You will also need to have redis server running locally. You will need docker running on your local machine and then run the following command:
+
+```sh
+docker run -p 6379:6379 -it redis/redis-stack-server:latest
```
notlee, line 35: We don't use docker for any of our other services, would an install via brew workout? This will avoid a contributor needing to also setup docker. https://redis.io/docs/install/install-redis/install-redis-on-mac-os/#installation

index.js
```diff
@@ -30,8 +31,22 @@ throng({
...
+  } else {
+    redisDb = await createClient({
+      url: process.env.REDIS_URL
+    });
+  }
+
+  redisDb.on('error', err => console.log('Redis Client Error', err)).connect();
+  
+  options.redisClient = redisDb;
+
```
notlee, line 49: We may want to wrap this whole thing in a try/catch block to avoid taking down the image service in the event there's an uncaught error in `createClient`?

lib/routes/v2/images.js
```diff
@@ -25,11 +25,19 @@ const {
...
 module.exports = app => {
+  const metrics = app.ft.metrics;
 
 	const requireValidSourceParam = sourceParam({
 		verifyUsingCmdb: false
 	});
 
+  // extract hostname from image url using regex
+  function extractHostname(url) {
+    const regex = /https?:\/\/(.+)\/.+/gm;
```
notlee, line 36: Instead of a regex, perhaps we could use [URL.hostname](https://nodejs.org/api/url.html#urlhostname)? It would be easier to read and perhaps more reliable. You would also have to handle a url which can't be parsed, with a try/catch, but that would be a good thing... we don't want to log anything for any invalid url

Looks like indentation is off. Could you match the rest of the file?

test/integration/v2/images-raw.test.js
```diff
@@ -990,24 +1005,39 @@ describe('GET /__origami/service/image/v2/images/raw…', function() {
 				});
 			});
 
-			describe('http', function() {
+			describe('http', async function() {
+        // await connectToDatabase();
```
notlee, line 1009: This is commented out and can be deleted. Does it depend on the `connectToDatabase` call in the previous test? It would be good to move this out into a `before` step, so it's clear that multiple tests need it

notlee (commented):
Looking like a nice solution! 😄 Just some small code quality suggestions/questions. Could you also check your indentation, and match the existing style?

test/integration/v2/images-raw.test.js
akomiqaia: yeah but before will re-initialise the database connections no? I mean it's fine but I thought just having one open client would be better

notlee (approved):

test/integration/v2/images-raw.test.js
notlee: You're thinking of `beforeEach`!

test/integration/v2/images-raw.test.js
```diff
@@ -1010,6 +1011,33 @@ describe('GET /__origami/service/image/v2/images/raw…', function() {
 				});
 			});
 
+			describe('http and https', async function() {
```
frshwtr, line 1014: nice testing here :)

frshwtr (approved):

notlee (approved):

PR #854, opened 2023-12-19
backstage: add tls for prod

akomiqaia:


lib/redis-client.js
```diff
@@ -2,6 +2,16 @@
...
-const redis = new Redis(process.env.REDIS_URL);
+
+const prodConfig = {
+  maxRetriesPerRequest: 1,
+  tls: {
+    rejectUnauthorized: false
+  },
+};
+
+const isProduction = process.env.REGION === 'EU' || process.env.REGION === 'US';
```
notlee, line 13: Are we able to use tls across all regions?

lib/redis-client.js
```diff
@@ -2,6 +2,16 @@
 
 require('dotenv').config();
 const Redis = require('ioredis');
-const redis = new Redis(process.env.REDIS_URL);
+
+const prodConfig = {
+  maxRetriesPerRequest: 1,
```
notlee, line 7: We may also want to reduce [connectTimeout](https://redis.github.io/ioredis/interfaces/CommonRedisOptions.html#connectTimeout) which is 10 seconds. Assuming `maxRetriesPerRequest` applies to making the connection, one retry alone would be 20 seconds.

lib/redis-client.js
akomiqaia: Not sure if our CI can establish TLS connection for testing. We could try but my assumption will be that we need TLS certificate for tests not to fail

lib/redis-client.js
akomiqaia: yeap. Do you think 3 seconds is good? Or too high/low?

lib/redis-client.js
notlee: I was thinking more about dev? That way we can use NODE_ENV instead, which feels a little bit more consistent and doesn't depend on listing out regions.

akomiqaia: yeah, but we don't have this defined in our secrets. and since this is a solution only for monitoring maybe we don't have to set a new env variable?

notlee: `NODE_ENV` is already set to `production` by default in heroku, and dev locally. If we could use tls across all heroku apps, then you could rely on `NODE_ENV`. This would also mean dev/qa are closer to production, and we wouldn't need to deploy to production to know for sure that this works

akomiqaia: but dev and qa will have it set to prod as well right? This means it will try to establish a TLS connection to a Redis instance that is not TLS-configured.

notlee: My question is, what's stopping us from using a tls connection for dev and qa?

notlee (approved):

PR #870, opened 2024-08-06
fix: add User-Agent to healthchecks

alexmuller:


notlee:
I've added to this to set `FTSystem/origami-image-service-v2` as the user agent when fetching images from source. This includes internal systems and, in the case of image sets hosted by the image service, requests to itself...

notlee (approved):
Little self approve here, huh 🤠

PR #886, opened 2024-08-13
Cut down the metrics sent by this service

sjparkinson:
This service sends a fair chunk of metrics, due to two things:

1. Fastly polls `__gtg` very frequently
2. It's an image proxy, so there are lots of client requests with _many_ different hosts

This change should ignore such requests while still retaining the most useful insights.

I've raised this as I've noticed a high max CPU and increased latency in the gateway collector just before this app does the daily dyno restart (though I might be miss attributing the cause here, it could easily also be preflight). I _think_ the JSON sent from this service gets bigger and bigger until the restart, as the number of host names accumulates with time.

As a note, we can always scale the collector (we're on tiny containers right now), but I figured this could also be a learning opportunity about the Node.js SDK and a chance to make your metrics more useful.


sjparkinson:
I've not seen this config used before to achieve this so it might be worth a little test.

opentelemetry.js
```diff
@@ -55,7 +55,20 @@ const sdk = new NodeSDK({
 
 	instrumentations: [
 		// Enable automatic instrumentation, see https://github.com/open-telemetry/opentelemetry-js-contrib/tree/main/metapackages/auto-instrumentations-node.
-		getNodeAutoInstrumentations(),
+		getNodeAutoInstrumentations({
+			'@opentelemetry/instrumentation-http': {
+				// Ignore the high volume requests to /__gtg
+				ignoreIncomingRequestHook: (request) => request.method?.toUpperCase() === 'HEAD',
```
sjparkinson, line 61: Based on https://github.com/getsentry/sentry-javascript/blob/615c670cfe283e77132339c3d9751060f30d3956/packages/node/src/integrations/http.ts#L117-L134.

It looks like in the previous commit, `request.url` is not actually a _URL_ object as I had assumed.

sjparkinson: Looks like this works as expected 👍 

I ran `ab -m HEAD -c "$(sysctl -n hw.ncpu)" -t 300 'https://origami-image-service-dev.herokuapp.com/__gtg'`, and checked the charts in Grafana. Nothing coming up in the server request metrics 🎉

sjparkinson:
Right, I'm happy this all works as expected.

Helpfully your dev instance is sending metrics into the deadletter workspace so I was able to test each hook using Apache Bench and check the hooks work as expected in Grafana.

j-dav (approved):
Looks good to me ⭐

sjparkinson:
For usage of these hooks, as it's not well documented, I found searching GitHub really helpful:

https://github.com/search?q=ignoreIncomingRequestHook+%28language%3ATypeScript+OR+language%3AJavaScript%29&type=code

j-dav:
> For usage of these hooks, as it's not well documented, I found searching GitHub really helpful:
> 
> https://github.com/search?q=ignoreIncomingRequestHook+%28language%3ATypeScript+OR+language%3AJavaScript%29&type=code

Deployed on prod

sjparkinson:
> Deployed on prod

Yeah, looks like it 😅

Screenshot 2024-08-19 at 14 23 42

Nice! Thanks for getting it released 🙇  This should make the generic dashboard a fair bit more useful for the service now too, https://grafana.ft.com/d/HzGduSwIz/node-js?orgId=1&var-workspace=I7o8Aa1Sz&var-job=origami-image-service-v2&var-environment=production&var-cloud_provider=heroku&var-cloud_region=All&var-http_route=All&var-http_errors=500&var-http_errors=502&var-http_errors=503&var-http_errors=504&var-http_errors=520&var-http_errors=530&from=now-15m&to=now.

PR #906, opened 2024-09-04
Replicate throng for worker processes with node cluster

j-dav:
- Replicated the throng single liner function
- lib is 4 years since last published
- Exactly same functionality, just easier for us to maintain


Sqyphen (approved):
Looks good, tests are all passing 👍🏼

sentayhuJo (approved):
Look good to me 👍

PR #935, opened 2024-09-11
[Update] Pin the file-type package version

Sqyphen:
Pin the file-type package to v16.5.4 as later versions required ESM module usage, rather than CommonJS


sentayhuJo (approved):
looks good to me !

PR #942, opened 2024-09-11
[Update] Node v18 to v20

Sqyphen:
Update uses of Node from v18 to v20


j-dav (approved):
LGTM

sentayhuJo (approved):
Looks good to me

PR #943, opened 2024-09-12
[Remove] Sentry from the image service.

sentayhuJo:
**What**
- Current subscription to Sentry will expire in Q4 2024. A reference to the Sentry SDK needs to be cleared.

**How**
- Remove all Sentry reference from code base
- Remove Heroku envs related to Sentry 
- Remove the project from Sentry, if possible. [access required]


j-dav (approved):
LGTM, I checked locally too, works fine

PR #960, opened 2024-10-03
[Remove] make-promises-safe package.

sentayhuJo:
**What**

- Remove `make-promises-safe` package

**How**
- Replace it with the native way of promise handling, `try-catch` and `await`


j-dav (approved):
LGTM. The error handling makes me think we need to create a ticket to implement the logger kit and have some concise error messaging/event names.

PR #961, opened 2024-10-03
[Remove] trace and clarify.

sentayhuJo:
**What** 

- Remove `trace`  and `clarify`

**How**
- Replaced by `--enable-source-maps` , which is Nodejs built in option. https://nodejs.org/docs/latest-v20.x/api/cli.html#--enable-source-maps


j-dav (approved):
Nice! Two less dependencies is always good

PR #962, opened 2024-10-04
[Remove]: heroku-node-settings

sentayhuJo:
**What**

- The image service is running on Node.js 20 and we need to move away from using `heroku-node-settings` tooling.
-  Heroku typically injects environment variables for configuration. we can access these directly in Node.js app without the need of this tool.

**How**
- Remove `heroku-node-setting` and `dotenv` from the project
- use `doppler run -c dev -- node....` to get all the secrete from doppler and run the app in development mode
- Use `require('./opentelemetry');` in `index.js` file , instead of using it in the `start` script


Sqyphen (approved):
Nice, good job on updating the readme too 👍🏼

PR #966, opened 2024-10-08
[Remove] More packages.

sentayhuJo:
**What** 

- Remove` path-exists`, `statuses`, `denodeify`,`start-server-and-test `and `supertest`


Sqyphen (approved):
👍🏼

